PR #1, opened 2020-09-24
Feature/logging

lecourtoisflo:


sources/CMakeLists.txt
```diff
@@ -7,21 +7,19 @@
...
-add_executable(DynaflowLauncher ${SOURCES} ${HEADERS})
+add_executable(DynaflowLauncher ${SOURCES})
 
 target_link_libraries(DynaflowLauncher
   PRIVATE
-  DynaflowLauncher::options
+  DynaflowLauncher::common
   Dynawo::dynawo_API_PAR
   Dynawo::dynawo_API_DYD
 )
```
gautierbureau, line 25: newline

gautierbureau (changes requested):

sources/common/include/Options.h
```diff
@@ -20,8 +20,8 @@
  * @brief Options header file
  */
 
-/// @brief Namespace for options library
-namespace options {
+/// @brief Namespace for common components to dynawo-launcher
+namespace common {
```
rosiereflo, line 24: common seems quite generic, maybe we should put DFL somewhere?

rosiereflo (changes requested):

gautierbureau:
General questions: we don't follow dynawo coding style in this project ? (directories not with capital letter, and structure sepration between src and include)

rosiereflo:
> General questions: we don't follow dynawo coding style in this project ? (directories not with capital letter, and structure sepration between src and include)

We should follow the same style in all our projects IMHO

lecourtoisflo:
> > General questions: we don't follow dynawo coding style in this project ? (directories not with capital letter, and structure sepration between src and include)
> 
> We should follow the same style in all our projects IMHO

It is not always respected in dynawo (util). As for include / src separation it is for public headers and sources. We don't see it in dynawo so I wonder.

lecourtoisflo:
> > > General questions: we don't follow dynawo coding style in this project ? (directories not with capital letter, and structure sepration between src and include)
> > 
> > 
> > We should follow the same style in all our projects IMHO
> 
> It is not always respected in dynawo (util). As for include / src separation it is for public headers and sources. We don't see it in dynawo so I wonder.

OK for name of he directories

sources/common/include/Options.h
lecourtoisflo: DFL ?

rosiereflo: DynaFlow-Launcher

sources/CMakeLists.txt
lecourtoisflo: OK

gautierbureau:
> > > General questions: we don't follow dynawo coding style in this project ? (directories not with capital letter, and structure sepration between src and include)
> > 
> > 
> > We should follow the same style in all our projects IMHO
> 
> It is not always respected in dynawo (util). As for include / src separation it is for public headers and sources. We don't see it in dynawo so I wonder.

I don't really mind if we do it but we have to be aware and make sure it is what we want. For folders with capitals it is only in the sources directory that is the rule, elsewhere where there is no cpp it is not. We can discuss to adapt them if need be

rosiereflo (approved):

gautierbureau (approved):

PR #2, opened 2020-09-25
Read network through DYNAWO API

lecourtoisflo:
depends on dynawo pull request 1146, deployed with c++11 enabled


sources/Common/include/Options.h
```diff
@@ -38,10 +38,10 @@ class Options {
    * representation after parsing
    */
   struct RuntimeConfiguration {
-    std::string programName;     ///< Name of the programm
-    std::string iidmPath;        ///< IIDM filepath ot process
-    std::string configPath;      ///< Launcher configuration filepath
-    std::string dynawoLogLevel;  ///< chosen log level
+    std::string programName;      ///< Name of the programm
```
rosiereflo, line 41: typo programm

sources/Common/src/Options.cpp
```diff
@@ -76,7 +76,7 @@ Options::basename(const std::string& filepath) {
 Options::Options() : desc_{}, config_{"", "", "", defaultLogLevel_} {
   desc_.add_options()("help,h", "Display help message")("log-level", po::value<ParsedLogLevel>(),
                                                         "Dynawo logger level (allowed values are ERROR, WARN, INFO, DEBUG): default is info")(
-      "iidm", po::value<std::string>(&config_.iidmPath)->required(), "IIDM file path to process")(
+      "iidm", po::value<std::string>(&config_.networkFilePath)->required(), "Network file path to process")(
```
rosiereflo, line 79: Would be good to have IIDM format named in the help message since this is the only format that we support so far

sources/Inputs/include/NetworkManager.h
```diff
@@ -0,0 +1,54 @@
...
+
+ private:
+  /**
+   * @brief Build node tree from data interface
+   */
+  void buildTree();
+
+ private:
+  boost::shared_ptr<DYN::DataInterface> interface_;        ///< data interface
+  std::map<Node::NodeId, boost::shared_ptr<Node>> nodes_;  ///< nodes representing the node tree
```
rosiereflo, line 50: Do we really need this to be sorted? Why not using unordered_map?

sources/Inputs/src/NetworkManager.cpp
```diff
@@ -0,0 +1,66 @@
...
+  buildTree();
+}
+
+void
+NetworkManager::buildTree() {
+  auto network = interface_->getNetwork();
+
+  const auto& voltageLevels = network->getVoltageLevels();
+  for (auto it_v = voltageLevels.begin(); it_v != voltageLevels.end(); ++it_v) {
+    if ((*it_v)->getVoltageLevelTopologyKind() == DYN::VoltageLevelInterface::BUS_BREAKER) {
```
rosiereflo, line 41: Will we add later NODE_BREAKER topology case?

sources/Inputs/include/Node.h
```diff
@@ -0,0 +1,89 @@
...
+   *
+   * @param node the new neighbour
+   */
+  void addNeighbour(boost::shared_ptr<Node> node) {
+    neighbours_.push_back(node);
+  }
+
+ private:
+  NodeId id_;                                        ///< node id
+  double voltageLevel_;                              ///< voltage level of the node
```
rosiereflo, line 84: This is the nominal voltage of the node if I am not mistaken, so let's update the comment

sources/Inputs/src/Node.cpp
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * @brief Node structure implementation file
+ *
+ */
+
+#include "Node.h"
+
+namespace dfl {
+namespace inputs {
+Node::Node(const NodeId& id, double voltageLevel) : id_(id), voltageLevel_{voltageLevel} {}
```
rosiereflo, line 22: voltageLevel->nominalVoltage

sources/main.cpp
```diff
@@ -25,12 +26,7 @@ main(int argc, char* argv[]) {
     }
     dfl::common::Log::init(options);
 
-    std::stringstream ss;
-    ss << "Hello World: ";
-    for (int i = 0; i < argc; i++) {
-      ss << argv[i] << " ";
-    }
-    LOG(info) << ss.str() << LOG_ENDL;
+    dfl::Context context(options.config().networkFilePath);
```
rosiereflo, line 29: No unit test? will they come later?

rosiereflo (changes requested):

sources/Common/include/Options.h
lecourtoisflo: OK

sources/Common/src/Options.cpp
lecourtoisflo: OK

sources/Inputs/include/NetworkManager.h
lecourtoisflo: insert is in O(log(N)) while for unordered_map it is O(N). For now, we access only for connection at the beginning but we go through the tree by walking through each node. For me, insert was the relevant operation, not the access

sources/Inputs/src/NetworkManager.cpp
lecourtoisflo: I used the algorithm in python script. I don't know if this is is still relevant

sources/Inputs/include/Node.h
lecourtoisflo: OK

sources/Inputs/src/Node.cpp
lecourtoisflo: OK

sources/main.cpp
lecourtoisflo: Context is not meant to have a unit test as it will be no different from launching the main. But network manager will have a unit test in future pull request

sources/Inputs/include/NetworkManager.h
rosiereflo: average insert for unordered_map is O(1), so if the order of the node is not relevant when visiting the tree unordered_map is definitively the best option

sources/Inputs/src/NetworkManager.cpp
rosiereflo: @cossartque @marcochiaramello any insight on this?

sources/Inputs/include/NetworkManager.h
rosiereflo: or maybe even a vector if research is not required

sources/Inputs/src/NetworkManager.cpp
cossartque: For now it is not necessary I guess. But at some point we might need it (maybe some automata will use a NODE_BREAKER topology). @marcochiaramello what do you think ?

marcochiaramello: For the moment we can rely only on BUS_BREAKER topology.

sources/Inputs/include/NetworkManager.h
lecourtoisflo: No, we need to perform connections between nodes, it will take too long with a vector

lecourtoisflo: OK for unordered_map

rosiereflo (approved):

sources/Inputs/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+)
+
+add_library(inputs STATIC ${SOURCES})
+
+target_include_directories(inputs
+PUBLIC
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+)
+target_link_libraries(inputs
+PUBLIC
```
gautierbureau, line 22: I don't know in this context but generally PRIVATE should be used by default unless there is a good reason for PUBLIC, reasons are listed here https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#transitive-usage-requirements

gautierbureau (changes requested):

CMakeLists.txt
```diff
@@ -67,17 +67,49 @@ cmake_dependent_option(DYNAFLOW_LAUNCHER_BUILD_TESTS
...
 ## Dynawo and sub-dependencies
 check_env_var(DYNAWO_HOME) # Correspond to deploy installation
-set(CMAKE_PREFIX_PATH APPEND "${DYNAWO_HOME}/share")
+list(APPEND CMAKE_PREFIX_PATH "${DYNAWO_HOME}/share")
 find_package(Dynawo 1.3.0 REQUIRED)
 #libXML
-set(CMAKE_PREFIX_PATH APPEND "${DYNAWO_HOME}/share/cmake")
+list(APPEND CMAKE_PREFIX_PATH "${DYNAWO_HOME}/share/cmake")
 find_package(libXML 0.2.4 REQUIRED)
+# XercesC (rewuired for libxml)
```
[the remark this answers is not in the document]
lecourtoisflo: OK

CMakeLists.txt
```diff
@@ -67,17 +67,49 @@ cmake_dependent_option(DYNAFLOW_LAUNCHER_BUILD_TESTS
...
+list(APPEND CMAKE_MODULE_PATH "${DYNAWO_HOME}/share/cmake")
+find_package(XercesC 3.2.2 REQUIRED)
```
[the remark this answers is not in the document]
lecourtoisflo: OK

sources/Inputs/CMakeLists.txt
lecourtoisflo: It is public because I want to link the transitive libraries at level above

gautierbureau: Public is ok but because you include #include <DYNDataInterface.h> in a .h and then you need transitivity, if it would only be included in .cpp it should then be private

gautierbureau (approved):

PR #3, opened 2020-09-28
Compute slack node using walkthrough network manager

lecourtoisflo:
Also rework test environment: now each sub-library has its own test suite


sources/Context.cpp
```diff
@@ -17,6 +17,29 @@
...
+
 namespace dfl {
-Context::Context(const std::string& networkFilepath) : networkManager_(networkFilepath) {}
+Context::Context(const std::string& networkFilepath) : networkManager_(networkFilepath) {
+  networkManager_.onNode(std::bind(&Context::processSlackNode, this, std::placeholders::_1));
+}
+
+void
+Context::processSlackNode(const boost::shared_ptr<inputs::Node>& node) {
+  if (!slack_node_) {
```
rosiereflo, line 29: I think it would be good to also write unit tests for those algorithms

tests/inputs/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
...
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, you can obtain one at http://mozilla.org/MPL/2.0/.
+# SPDX-License-Identifier: MPL-2.0
+#
+
+DEFINE_TEST(TestNetworkManager INPUTS)
+target_link_libraries(TestNetworkManager DynaflowLauncher::inputs)
+
+set_property(TEST INPUTS.TestNetworkManager PROPERTY ENVIRONMENT IIDM_XML_XSD_PATH="${DYNAWo_HOME}/share/iidm/xsd/")
```
rosiereflo, line 13: capital O in DYNAWO_HOME

rosiereflo (changes requested):
very elegant solution : )

sources/Context.cpp
rosiereflo: Solution: Algorithms will be written in a new library that will be unit-tested, and Context will simply aggregate all the algorithms from the library. Will be done in a future PR.

tests/inputs/CMakeLists.txt
lecourtoisflo: OK

sources/Context.cpp
lecourtoisflo: Algorithms will be put in a separated ibrary to be unit-tested

rosiereflo (approved):
OK on my side, I will let Marco take a look at the algorithm itself

marcochiaramello (approved):
ok for me!

gautierbureau (approved):

PR #5, opened 2020-09-29
Add dictionnary to manage language of the logs

lecourtoisflo:


README.md
```diff
@@ -1 +1,29 @@
...
+  * en_GB (English)
+
+Then build the project in the build directory:
+
+`cmake --build . --target install`
+
+## Run
+To run DynaFlow launcher, the following environment variables must be defined:
+* LD_LIBRARY_PATH: must contain the path to the deployed shared libraries of dynawo
+* IIDM_XML_XSD_PATH: must be the path to xsd of dynawo (generally ${DYNAWO_HOME}/share/iidm/xsd)
```
rosiereflo, line 22: Maybe we could have a default value for this one based on DYNAWO_HOME?

sources/Common/include/Dico.h
```diff
@@ -0,0 +1,79 @@
...
+// Copyright (c) 2020, RTE (http://www.rte-france.com)
+// See AUTHORS.txt
+// All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at http://mozilla.org/MPL/2.0/.
+// SPDX-License-Identifier: MPL-2.0
+//
+
+#pragma once
```
rosiereflo, line 11: Cannot we reuse dynawo classes for this?

sources/Common/scripts/generateMessages.py
```diff
@@ -0,0 +1,170 @@
...
+
+import sys
+import re
+import os
+
+
+# @brief Dictionnary definition for message keys
+#
+class Dictionnary:
+
```
rosiereflo, line 20: Same question as above

sources/Context.h
```diff
@@ -45,6 +45,13 @@ class Context {
    */
   void process();
 
+ private:
+  /// @brief Slack node origin
+  enum class SlackNodeOrigin {
+    FILE = 0,  ///< Slack node comes from network input file
+    ALGORITHM  ///< Slack node is computed via automatic algorithm
```
rosiereflo, line 52: Maybe we should add a UNKNOWN value to cover potential cases for which we do not assign the origin?

sources/main.cpp
```diff
@@ -8,25 +8,58 @@
...
 
     if (!options.parse(argc, argv)) {
-      LOG(error) << options << LOG_ENDL;
+      LOG(info) << options << LOG_ENDL;
       return 0;
     }
     dfl::common::Log::init(options);
 
+    std::string locale = getMandatoryEnvVar("DYNAFLOW_LAUNCHER_LOCALE");
+    std::string dictDir = getMandatoryEnvVar("DYNAFLOW_LAUNCHER_DICTS");
```
rosiereflo, line 48: This mandatory environment variable is not specified in the README

rosiereflo (changes requested):

README.md
lecourtoisflo: This is a runtime environment variable used by Dynawo. Settings a variable environment in code seems unwise to me

sources/Common/include/Dico.h
lecourtoisflo: I'm not sure of the behaviour of the singleton IoDicos if we use it directly

sources/Context.h
lecourtoisflo: We always assign the origin: it depends only of the configuration (with at worst has a default value) and the iidm file, which is required

sources/main.cpp
lecourtoisflo: OK

README.md
rosiereflo: I mean a default value in the code if this variable or the one below is not defined

sources/Context.h
rosiereflo: This is true today, what about tomorrow?

lecourtoisflo: I don't think the algorithms will change a lot. Besides, for now, we use always the algorithm case since the xml element in the iidm doesn't exist

rosiereflo (approved):

gautierbureau (approved):

PR #6, opened 2020-09-30
Add algo library

lecourtoisflo:
Put algorithms in their own library in order to test them independently

### DIFF
--- a/sources/Algo/src/Algo.cpp
+++ b/sources/Algo/src/Algo.cpp
@@ -1,0 +1,40 @@
+//
+// Copyright (c) 2020, RTE (http://www.rte-france.com)
+// See AUTHORS.txt
+// All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at http://mozilla.org/MPL/2.0/.
+// SPDX-License-Identifier: MPL-2.0
+//
+
+/**
+ * @file  Algo.cpp
+ *
+ * @brief Dynaflow launcher algorithms implementation file
+ *
+ */
+
+#include "Algo.h"
+
+#include <tuple>
+
+namespace dfl {
+namespace algo {
+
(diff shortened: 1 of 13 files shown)


sources/Algo/src/Algo.cpp
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+#include "Algo.h"
+
+#include <tuple>
+
+namespace dfl {
+namespace algo {
+
+SlackNodeAlgorithm::SlackNodeAlgorithm(std::shared_ptr<inputs::Node>& slackNode) : slackNode_{slackNode} {}
```
rosiereflo, line 25: Do we plan to add all algorithms into the same file?

rosiereflo (changes requested):
Just one question

sources/Inputs/include/NetworkManager.h
```diff
@@ -72,9 +74,9 @@ class NetworkManager {
   void buildTree();
 
  private:
-  boost::shared_ptr<DYN::DataInterface> interface_;                  ///< data interface
-  std::unordered_map<Node::NodeId, boost::shared_ptr<Node>> nodes_;  ///< nodes representing the node tree
-  std::vector<ProcessNodeCallback> nodesCallbacks_;                  ///< list of callback or nodes
+  boost::shared_ptr<DYN::DataInterface> interface_;      ///< data interface
```
gautierbureau, line 77: why keep boost::shared_ptr here ? if not useless include at the top of the file

gautierbureau (approved):

sources/Algo/src/Algo.cpp
lecourtoisflo: Yes. For me we can put all algorithms in the same file, since they seems small

sources/Inputs/include/NetworkManager.h
lecourtoisflo: This is the returned value of the Dynawo API

sources/Algo/src/Algo.cpp
rosiereflo: OK, let's keep this for now but if the file becomes too large we will need to split it

rosiereflo (approved):

PR #8, opened 2020-10-01
Add job file writer

lecourtoisflo:
Also add install run script


README.md
```diff
@@ -21,6 +21,7 @@ To run DynaFlow launcher, the following environment variables must be defined:
 * IIDM_XML_XSD_PATH: must be the path to xsd of dynawo (generally ${DYNAWO_HOME}/share/iidm/xsd depending on the deployement of Dynawo)
 * DYNAFLOW_LAUNCHER_LOCALE: the locale dictionnary to use. The supported values are the same as for the build. The runtime value can be different from the one used in compilation but all keys must be the same in both files and must use the same number of arguments
 * DYNAFLOW_LAUNCHER_DICTS: the directory where the dictionnaries are located
+* DYNAFLOW_LAUNCHER_PAR: the directory where the constant parameter files are located
```
rosiereflo, line 24: This shouldn't be editable by the user

rosiereflo: or even visible

sources/Common/include/Options.h
```diff
@@ -44,6 +45,15 @@ class Options {
     std::string dynawoLogLevel;   ///< chosen log level
   };
 
+  /**
+   * @brief type of request for dynaflow launcher
+   */
+  enum class Request {
+    NORMAL = 0,  ///< Normal run of launcher
+    HELP,        ///< help display is requested
+    VERSION      ///< versio display is requested
```
rosiereflo, line 54: typo : version

sources/Common/include/Options.h
```diff
@@ -44,6 +45,15 @@ class Options {
     std::string dynawoLogLevel;   ///< chosen log level
   };
 
+  /**
+   * @brief type of request for dynaflow launcher
+   */
+  enum class Request {
+    NORMAL = 0,  ///< Normal run of launcher
```
rosiereflo, line 52: NORMAL=> RUN_SIMULATION

sources/Context.cpp
```diff
@@ -51,4 +61,23 @@ Context::process() {
   LOG(info) << MESS(SlackNode, slackNode_->id, static_cast<unsigned int>(slackNodeOrigin_)) << LOG_ENDL;
 }
 
+void
+Context::exportOutputs() {
+  LOG(info) << MESS(ExportInfo, basename_) << LOG_ENDL;
+
+  // Job
+  jobWriter_ = std::unique_ptr<outputs::Job>(new outputs::Job(outputs::Job::JobDefinition(config_.outputDir(), basename_, dynawLogLevel_)));
```
rosiereflo, line 69: not sure why we write a job when the input of the Simulation class in Dynawo is directly a jobEntry

etc/Network.par
```diff
@@ -0,0 +1,37 @@
...
+    simulation tool for power systems.
+-->
+<parametersSet xmlns="http://www.rte-france.com/dynawo">
+  <set id="Network">
+    <par type="DOUBLE" name="capacitor_no_reclosing_delay" value="300"/>
+    <par type="DOUBLE" name="dangling_line_currentLimit_maxTimeOperation" value="90"/>
+    <par type="DOUBLE" name="line_currentLimit_maxTimeOperation" value="90"/>
+    <par type="DOUBLE" name="load_Tp" value="90"/>
+    <par type="DOUBLE" name="load_Tq" value="90"/>
+    <par type="DOUBLE" name="load_alpha" value="1"/>
```
cossartque, line 21: @marcochiaramello do you think we should put alpha and beta to zero to have PQ load when using the cpp model or keep 1 and 2 ?

rosiereflo (changes requested):
2 main things here:
- I don't think we need to write a job on the disk, since we can call a dynawo simulation directly with a jobEntry
- User shouldn't have to configure anything related to constant files

etc/solver.par
```diff
@@ -0,0 +1,26 @@
...
+    This Source Code Form is subject to the terms of the Mozilla Public
+    License, v. 2.0. If a copy of the MPL was not distributed with this
+    file, you can obtain one at http://mozilla.org/MPL/2.0/.
+    SPDX-License-Identifier: MPL-2.0
+
+    This file is part of Dynawo, an hybrid C++/Modelica open source time domain
+    simulation tool for power systems.
+-->
+<parametersSet xmlns="http://www.rte-france.com/dynawo">
+  <set id="SimplifiedSolver">
```
cossartque, line 15: Maybe you can put more parameters like in the set id="0" of dynawo-rte/nrt/data/PtFige-Nancy-DynaFlow/20170619_0955/tFin/fic_PAR.xml

README.md
rosiereflo: (this is also true for DYNAFLOW_LAUNCHER_DICTS)

cossartque (changes requested):

sources/Common/include/Options.h
lecourtoisflo: OK

sources/Common/include/Options.h
lecourtoisflo: OK

scripts/run_install.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+INSTALL=`dirname "$0"`
```
gautierbureau, line 2: $(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd) is better

scripts/run_install.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
gautierbureau, line 1: dynaflow-launcher.sh would be a better name

sources/Context.cpp
lecourtoisflo: > 2 main things here:
> 
>     * I don't think we need to write a job on the disk, since we can call a dynawo simulation directly with a jobEntry
> 
>     * User shouldn't have to configure anything related to constant files

For a first draft, we create the input files to be able to test that we provide the correct informations to dynawo siulation

README.md
lecourtoisflo: This location is not the same if we use the installation or if we use the normal binary so it must be editable. DYNAFLOW_LAUNCHER_DICTS and DYNAFLOW_LAUNCHER_PAR may be fused in DYNAFLOW_LAUNCHER_RESOURCES

sources/Context.cpp
rosiereflo: This can be tested directly on the jobEntry (and displayed if needed). for others files we can have a first shot with files written on the disk, but for the jobs it seems useless

README.md
rosiereflo: I would propose DYNAWFLOW_LAUNCHER_INSTALL?

sources/Outputs/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+
+target_include_directories(outputs
+PUBLIC
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+)
+
+target_set_warnings(outputs ENABLE ALL AS_ERROR ALL DISABLE Annoying) # Helper that can set default warning flags for you
+target_link_libraries(outputs
+PUBLIC
+DynaFlowLauncher::common
```
gautierbureau, line 24: I don't see the need to link to this for the moment

sources/Context.cpp
rosiereflo: (sorry if I wasn't clear on this topic)

etc/solver.par
lecourtoisflo: OK

gautierbureau (changes requested):

etc/Network.par
marcochiaramello: I'd put alpha and beta equal to zero and load_isRestorative equals to false, so that we have a pure PQ injection when we are using a c++ model

README.md
lecourtoisflo: DYNAWFLOW_LAUNCHER_INSTALL correspond to CMAKE_PREFIX_INSTALL

etc/Network.par
lecourtoisflo: OK

README.md
rosiereflo: yes, and user shouldn't have to understand what is put inside this, so this is the only path (along with a distribution) that we can ask him to provide

sources/Context.cpp
lecourtoisflo: For me, we have to perform the same tests for all files, even if it is less relevant for the job file. That way all inputs of dynawo simulation are tested the same way in draft. JOB entry generation will be done when the simulation will be launched from inside the launcher

rosiereflo: I don't think we need a step in which dynawo is called as a blackbox since we already know this works.
@marcochiaramello @cossartque what do you think? I strongly think that we can already go to a version in which the simulation is directly launched from inside the tool as this is something that was already tested extensively.

cossartque: I agree, I also think that we can launch the simulation directly from inside the tool.

marcochiaramello: agreed

README.md
lecourtoisflo: OK

sources/Context.cpp
lecourtoisflo: OK, I'll reverse this in another pull request

scripts/run_install.sh
lecourtoisflo: Why ?

scripts/run_install.sh
lecourtoisflo: OK

sources/Context.cpp
rosiereflo: OK thanks! and sorry for the misunderstanding

rosiereflo (approved):

sources/Outputs/CMakeLists.txt
lecourtoisflo: OK

cossartque (approved):

marcochiaramello (approved):

scripts/run_install.sh
gautierbureau: works with relative and absolute path

lecourtoisflo: OK

scripts/dynaflow-launcher.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+INSTALL=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+export DYNAWO_HOME=$INSTALL/external/dynawo
+export LD_LIBRARY_PATH=$DYNAWO_HOME/lib:$LD_LIBRARY_PATH
+export IIDM_XML_XSD_PATH=$DYNAWO_HOME/share/iidm/xsd
+export DYNAFLOW_LAUNCHER_LOCALE=en_GB
+export DYNAFLOW_LAUNCHER_DICTS=$INSTALL/etc
+export DYNAFLOW_LAUNCHER_PAR=$INSTALL/etc
+
+$INSTALL/bin/DynaFlowLauncher $@
```
PerninGuillaume, line 10: The $@ should be quoted.

lecourtoisflo: There are not quotes in the equivalent in Dynawo

PerninGuillaume: I know but there should, it is always best to quote the variables (that is what I learnt at school anyway) because of the unintentional expansion that could occur (cf the first answer to the question here : [https://unix.stackexchange.com/questions/129072/whats-the-difference-between-and](url)

gautierbureau (approved):

PR #9, opened 2020-10-02
Add algoritm to compute connexity

lecourtoisflo:


sources/Context.cpp
```diff
@@ -51,14 +51,30 @@ Context::Context(const std::string& networkFilepath, const std::string& configFi
...
+
+bool
 Context::process() {
   // Process all algorithms on nodes
   networkManager_.walkNodes();
 
   LOG(info) << MESS(SlackNode, slackNode_->id, static_cast<unsigned int>(slackNodeOrigin_)) << LOG_ENDL;
+
+  if (!checkConnexity()) {
+    LOG(error) << MESS(ConnexityError, slackNode_->id) << LOG_ENDL;
```
rosiereflo, line 73: If the selection of the slack node is automatic, maybe we should look only in the main connex component? Otherwise the user is stuck.

sources/Context.h
```diff
@@ -72,7 +84,8 @@ class Context {
 
   std::shared_ptr<inputs::Node> slackNode_;  ///< computed slack node
   SlackNodeOrigin slackNodeOrigin_;          ///< slack node origin
+  std::vector<std::shared_ptr<inputs::Node>> mainConnexNodes_;
```
rosiereflo, line 87: Missing comment

sources/main.cpp
```diff
@@ -74,7 +74,9 @@ main(int argc, char* argv[]) {
 
     dfl::Context context(config.networkFilePath, config.configPath, config.dynawoLogLevel, parFilesDir.generic_string());
 
-    context.process();
+    if (!context.process()) {
+      return -1;
```
rosiereflo, line 78: Maybe we should use EXIT_FAILURE?

tests/algo/TestAlgo.cpp
```diff
@@ -90,10 +87,40 @@ TEST(SlackNodeAlgo, equivalent) {
...
+  nodes[6]->neighbours.push_back(nodes[5]);
+
+  dfl::algo::ConnexityAlgorithm::ConnexGroup main;
+  dfl::algo::ConnexityAlgorithm algo(main);
+
+  std::for_each(nodes.begin(), nodes.end(), algo);
+
+  ASSERT_EQ(4, main.size());
+  std::vector<dfl::inputs::Node::NodeId> nodeids_main;
+  std::for_each(main.begin(), main.end(), [&nodeids_main](const std::shared_ptr<dfl::inputs::Node>& node) { nodeids_main.push_back(node->id); });
```
rosiereflo, line 123: I would like to see:
- unit test with same number of nodes on right or left (@marcochiaramello is there a property that we could use as a tie -breaker in this case?
- a scaling unit test with a large complete graph

rosiereflo (changes requested):

sources/Context.cpp
cossartque: I agree that the automatic slack node calculation should be in the main cc only

sources/Algo/include/Algo.h
```diff
@@ -42,10 +52,72 @@ class SlackNodeAlgorithm {
...
   */
-  void operator()(const std::shared_ptr<inputs::Node>& node);
+  void operator()(const NodePtr& node);
+
+ private:
+  NodePtr& slackNode_;  ///< The slack node to update
+};
+
+/**
+ * @brief Algorithm to determine connexity
```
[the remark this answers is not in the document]
lecourtoisflo: OK

sources/Context.cpp
lecourtoisflo: This would require to change the architecture idea, since slack node and main connex component are determined in the same time. Feasible but not in the spirit of the beginning.

Besides, isn't that weird that the slack node in the whole network is not in the main connex component to begin with ? Are we sure we don't want to notify the user of that ?

sources/Context.h
lecourtoisflo: OK

sources/main.cpp
lecourtoisflo: OK

sources/Context.cpp
rosiereflo: for the first point, then we should adapt the architecture if required.
For the second point, no it is not weird, we can have 2 sub network working in parallel in a simulation, it's just that dynawo is not able to handle that today.

tests/algo/TestAlgo.cpp
lecourtoisflo: - what do you mean same number of nodes on right or left ?
- As I don't understand quite how the IIDM is built, I need a translation of a "large complete graph" in terms of nodes (id + nominal voltage)

rosiereflo: - I mean 2 or more subgraphs with the same number of node
- I was thinking about a theoretical graph that has a large number of nodes and connections (a complete graph is a graph for which all nodes are connected to all others nodes, but I do agree it is not really representative in this case). We could define a nrt with the IIDM of one of the testcases of Marco and Quentin so that is is more meaningful. Let's drop this one for now.

rosiereflo: Also could you please add later the possibility to display the wall time of the algorithms?

sources/Context.cpp
lecourtoisflo: After discussion with @cossartque, this is what we'll do:
- if slack node is given in the IIDM file but not in the main connex component, send error
- if slack node is computed automatically, compute it first on the whole network then, if not in the main component, computes it only on the main component. That way, we avoid going through nodes of the network twice in a lot of cases but only in a worst case

rosiereflo: why not using the result of the connex component research during the slack node research?

lecourtoisflo: They are done on the same time. Each algorithm is computed node by node, meaning that we don't know the main connex component when we start the algorithm of slack node

rosiereflo: since everything is done on the main connex component, we might need later to first find it and then call algorithms only on it (and more later on all the connex components), but for today let's do this

rosiereflo (approved):

cossartque (approved):

PerninGuillaume (approved):
I approve these changes and liked the use of functors to make the code clean.

gautierbureau (approved):

PR #16, opened 2020-10-08
Add par file  writer

lecourtoisflo:


sources/Outputs/src/Constants.h
```diff
@@ -0,0 +1,38 @@
...
+namespace dfl {
+namespace outputs {
+namespace constants {
+const std::string loadParId("GenericRestorativeLoad");              ///< PAR id common to all loads
+const std::string signalNParId("SignalN");                          ///< PAR id for signalN
+const std::string diagramFileSuffixExt("_Diagram.txt");             ///< Basename for QTables diagrams
+const std::string diagramMaxTableSuffix("_tableqmax");              ///< Suffix for the table name for qmax in diagram file
+const std::string diagramMinTableSuffix("_tableqmin");              ///< Suffix for the table name for qmin in diagram file
+const std::string signalNGeneratorParId("signalNGenerator");        ///< PAR id for generators using signal N
+const std::string impSignalNGeneratorParId("impSignalNGenerator");  ///< PAR id for generators using signal N
```
PerninGuillaume, line 32: Same comments for the two variables, intentional ?

PerninGuillaume (approved):

sources/Outputs/src/Constants.h
lecourtoisflo: No, it shoudl be "PAR id for generators using signal N with impendance" for the second one. correction will be done

sources/Outputs/src/Par.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+  exporter.exportToFile(collection, def_.filepath);
+}
+
+void
+Par::updateSignalNGenerator(boost::shared_ptr<parameters::ParametersSet> set) {
+  set->addParameter(helper::buildParameter("generator_KGover", 1.));
+  set->addParameter(helper::buildParameter("generator_QMin", -constants::qValueMax));
+  set->addParameter(helper::buildParameter("generator_QMax", constants::qValueMax));
+  set->addParameter(helper::buildParameter("generator_PMin", -constants::qValueMax));
```
rosiereflo, line 83: Please rename the constant if it is also used for active power

tests/main/res/config_launch_infinite.json
```diff
@@ -0,0 +1,6 @@
+{
+  "dfl-config": {
+    "OutputDir": "results/launch_infinite",
```
rosiereflo, line 3: I didn't see where this new file was used?

rosiereflo (changes requested):

tests/main/res/config_launch_infinite.json
lecourtoisflo: In tests/main we add a test which has it in input configuration

sources/Outputs/src/Par.cpp
lecourtoisflo: OK

sources/Outputs/src/Constants.h
```diff
@@ -31,7 +31,7 @@ const std::string diagramMinTableSuffix("_tableqmin");              ///< Suffix
 const std::string signalNGeneratorParId("signalNGenerator");        ///< PAR id for generators using signal N
 const std::string impSignalNGeneratorParId("impSignalNGenerator");  ///< PAR id for generators using signal N with impendance
 
-constexpr double qValueMax = std::numeric_limits<double>::max();  ///< Maximum value for reactive power, meaning infinite
+constexpr double powerValueMax = std::numeric_limits<double>::max();  ///< Maximum value for reactive power, meaning infinite
```
rosiereflo, line 34: Please update the comment

rosiereflo (changes requested):

sources/Outputs/src/Constants.h
lecourtoisflo: OK

sources/Outputs/src/Par.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+  set->addParameter(helper::buildParameter("generator_KGover", 1.));
+  if (def.model == algo::GeneratorDefinition::ModelType::WITH_IMPEDANCE_DIAGRAM_PQ_SIGNALN ||
+      def.model == algo::GeneratorDefinition::ModelType::WITH_IMPEDANCE_SIGNALN) {
+    updateCouplingParameters(set);
+  }
+
+  set->addParameter(helper::buildParameter("generator_tFilter", 0.001));
+  // Qmax and QMin are determined in dynawo according to reactive capabilities curves and min max
+  set->addParameter(helper::buildParameter("generator_QMin0", def.qmin - 1));
```
marcochiaramello, line 146: could you add a comment here about def.qmin -1 and def.qmax + 1, so that we do not forget why we do that? the answer is that we need a small numerical tolerance in case the starting point of the reactive injection is exactly on the limit of the reactive capability curve

marcochiaramello (changes requested):

sources/Outputs/src/Par.cpp
lecourtoisflo: OK

gautierbureau (approved):

rosiereflo (approved):

marcochiaramello (approved):

cossartque (approved):

PR #19, opened 2020-10-12
Run simulation

lecourtoisflo:
Update log appenders
Improve logs and comments


etc/DFLMessages_en_GB.dic
```diff
@@ -14,3 +14,4 @@ ExportInfo                    =     Exporting outputs files for %1%
 SlackNode                     =     Slack node of id %1% found with origin %2% (0=file, 1=algorithm)
 ConnexityError                =     Slack node of id %1% not present in main connex component
 ConnexityErrorReCompute       =     Slack node of id %1% not present in main connex component: compute slack node only in main connex component
+SimulateInfo                  =     Creating simulation for %1%
```
rosiereflo, line 17: I would propose: "initializing %1% simulation"

sources/Inputs/src/NetworkManager.cpp
```diff
@@ -86,6 +91,7 @@ NetworkManager::buildTree() {
     if (bus1 && bus2) {
       nodes_[bus1->getID()]->neighbours.push_back(nodes_[bus2->getID()]);
       nodes_[bus2->getID()]->neighbours.push_back(nodes_[bus1->getID()]);
+      LOG(debug) << "Node " << bus1->getID() << " connected to " << bus2->getID() << LOG_ENDL;
```
rosiereflo, line 94: maybe we could add the name of the line/2wt/3wt in the message?

sources/Outputs/src/Job.cpp
```diff
@@ -46,7 +46,7 @@ namespace dfl {
 namespace outputs {
 
 const std::chrono::seconds Job::timeStart_{0};
-const std::chrono::seconds Job::durationSimu_{30};
+const std::chrono::seconds Job::durationSimu_{100};
```
rosiereflo, line 49: Was this discussed with Marco and Quentin?

sources/Outputs/src/Job.cpp
```diff
@@ -119,6 +119,18 @@ Job::writeOutputs() {
   appender->setFilePath("dynawo.log");
   appender->setLvlFilter(def_.dynawoLogLevel);
   log->addAppenderEntry(appender);
+
+  appender = job::AppenderEntryFactory::newInstance();
+  appender->setTag("COMPILE");
```
rosiereflo, line 124: I don't think we should dump those files by default

rosiereflo (changes requested):

etc/DFLMessages_en_GB.dic
lecourtoisflo: OK

sources/Inputs/src/NetworkManager.cpp
lecourtoisflo: OK

sources/Outputs/src/Job.cpp
lecourtoisflo: Yes. this modification is due to comparaison between results on IEEE14 python and DFL. 30s was not enough to reach the asymptot

sources/Outputs/src/Job.cpp
lecourtoisflo: I put them because I rememeber that for the first test I performed with DFL, dynawo log was not enought. But maybe it was just the modeler log that is required

rosiereflo: Both of them are debug logs, so I am not sure we should use them by default. Maybe we could add them with a user option? @marcochiaramello any input on this?

gautierbureau: And COMPILE is only here to get logs if you use models compiled on the fly by dynawo whihch should never happen for dynaflo as we will only use black box models to speed up the simulation

gautierbureau (approved):

sources/Outputs/src/Job.cpp
lecourtoisflo: what about MODELER log ?

rosiereflo: MODELER is really a debug log that provides the index of each model for example

lecourtoisflo: OK, I remove it as well

rosiereflo (approved):

marcochiaramello (approved):

PerninGuillaume (approved):

PR #21, opened 2020-10-13
Feature/coverage

lecourtoisflo:


tests/inputs/TestNode.cpp
```diff
@@ -0,0 +1,29 @@
...
+//
+
+#include "Node.h"
+#include "Tests.h"
+
+TEST(TestNode, base) {
+  dfl::inputs::Node node("0", 0.0);
+  dfl::inputs::Node node0("0", 10.);
+  dfl::inputs::Node node2("1", 4.5);
+
```
PerninGuillaume, line 18: Why the names node,node0 and then node2 ?

lecourtoisflo: node0 because it has the same node id as "node". node2 is an artifact of the draft if the test where there were node, node1n and node2

rosiereflo (approved):

PerninGuillaume (approved):

gautierbureau (approved):

PR #23, opened 2020-10-15
Improve doxygen documentation and comments

lecourtoisflo:


sources/Common/include/Dico.h
```diff
@@ -8,6 +8,11 @@
 // SPDX-License-Identifier: MPL-2.0
 //
 
+/**
+ * @file Dico.h
+ * @brief Dictionnary manager header file
```
rosiereflo, line 13: typo: Dictionary

sources/Common/include/Message.hpp
```diff
@@ -8,6 +8,11 @@
 // SPDX-License-Identifier: MPL-2.0
 //
 
+/**
+ * @file Message.hpp
+ * @brief Dictionnary message file
```
rosiereflo, line 13: same as above

sources/Common/CMakeLists.txt
```diff
@@ -28,8 +28,6 @@ check_env_var(DYNAFLOW_LAUNCHER_LOCALE)
 # We use a target to ensure it will be executed at each build
 add_custom_target(dicokeys
 BYPRODUCTS ${CMAKE_CURRENT_SOURCE_DIR}/src/DicoKeys.cpp ${CMAKE_CURRENT_SOURCE_DIR}/include/DicoKeys.h
-COMMAND rm -f ${CMAKE_CURRENT_SOURCE_DIR}/src/DicoKeys.cpp
```
rosiereflo, line 31: Useless as the file is systematically erase by the script right?

sources/Common/src/Dico.cpp
```diff
@@ -8,6 +8,11 @@
 // SPDX-License-Identifier: MPL-2.0
 //
 
+/**
+ * @file Dico.cpp
+ * @brief Dictionnary manager impelmentation file
```
rosiereflo, line 13: same typo

rosiereflo (changes requested):

sources/Common/include/Dico.h
lecourtoisflo: OK

sources/Common/include/Message.hpp
lecourtoisflo: OK

sources/Common/CMakeLists.txt
lecourtoisflo: Yes, it is always recreated even if it already exists

sources/Common/src/Dico.cpp
lecourtoisflo: OK

rosiereflo (approved):

gautierbureau (approved):

sources/Outputs/src/Constants.h
```diff
@@ -22,14 +22,15 @@
...
+/// @brief Namespace for constant variables common to all writers
 namespace constants {
-const std::string loadParId("GenericRestorativeLoad");              ///< PAR id common to all loads
-const std::string signalNParId("SignalN");                          ///< PAR id for signalN
-const std::string diagramFileSuffixExt("_Diagram.txt");             ///< Basename for QTables diagrams
-const std::string diagramMaxTableSuffix("_tableqmax");              ///< Suffix for the table name for qmax in diagram file
-const std::string diagramMinTableSuffix("_tableqmin");              ///< Suffix for the table name for qmin in diagram file
-const std::string signalNGeneratorParId("signalNGenerator");        ///< PAR id for generators using signal N
-const std::string impSignalNGeneratorParId("impSignalNGenerator");  ///< PAR id for generators using signal N with impendance
+const std::string loadParId{"GenericRestorativeLoad"};              ///< PAR id common to all loads
```
PerninGuillaume, line 27: What is the difference between the initialization between () and {} ?

lecourtoisflo: () uses the constructor, {} uses the initializer list, which sometimes is better. But in that case, the rationnal for {} is because if we use (), doxygen think that they are functions and if we use {}, it understands correctly that they are variables

PerninGuillaume (approved):

Issue #31, opened 2020-10-20
Logs won't appear in DFL log file after dynawo simulation has ended


rosiereflo:

rosiereflo:
This one is dependent on ticket 1215 in dynawo repository

PR #40, opened 2020-10-21
#31 Append to log file after simulation ended

lecourtoisflo:
Linked to Dynawo PR #1218

### DIFF
--- a/sources/Common/src/Log.cpp
+++ b/sources/Common/src/Log.cpp
@@ -25,7 +25,7 @@ namespace common {
 const char* Log::dynaflowLauncherLogTag = "DYNAFLOW_LAUNCHER";
 
 void
-Log::init(const common::Options& options, const std::string& outputDir) {
+Log::init(const common::Options& options, const std::string& outputDir, bool continuePreviousLog) {
   using DYN::Trace;
   auto& config = options.config();
 
@@ -42,6 +42,8 @@ Log::init(const common::Options& options, const std::string& outputDir) {
   appender.setShowLevelTag(true);
   appender.setSeparator(" ");
   appender.setShowTimeStamp(true);
+  appender.setAppend(continuePreviousLog);
+  appender.setAppend(continuePreviousLog);
   appender.setTimeStampFormat("%Y-%m-%d %H:%M:%S");
 
   appenders.push_back(appender);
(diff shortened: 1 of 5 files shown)


PerninGuillaume (approved):

sources/main.cpp
```diff
@@ -98,9 +98,12 @@ main(int argc, char* argv[]) {
     }
 
     context.exportOutputs();
-
     context.execute();
 
+    dfl::common::Log::init(options, config.outputDir(), true);
+
+    LOG(info) << MESS(SimulationEnded, context.basename()) << LOG_ENDL;
```
gautierbureau, line 105: This means that we don't give any information in the console when simulation is ended ?

lecourtoisflo: Depends on Dynawo implementation but not in DFL

rosiereflo: There is another ticket about ease of use and logs of DFL

gautierbureau: ok because we add the experience with dynawo never saying Simulation finished OK or NOK in the console before and it is better to have a small message for the user at the end of the simulation in the console. if it is addressed later on then no problem.

gautierbureau (approved):

sources/Common/src/Log.cpp
```diff
@@ -42,6 +42,8 @@ Log::init(const common::Options& options, const std::string& outputDir) {
   appender.setShowLevelTag(true);
   appender.setSeparator(" ");
   appender.setShowTimeStamp(true);
+  appender.setAppend(continuePreviousLog);
+  appender.setAppend(continuePreviousLog);
```
rosiereflo, line 46: Why do it twice?

sources/main.cpp
```diff
@@ -64,7 +64,7 @@ main(int argc, char* argv[]) {
 
     auto& runtimeConfig = options.config();
     dfl::inputs::Configuration config(runtimeConfig.configPath);
-    dfl::common::Log::init(options, config.outputDir());
+    dfl::common::Log::init(options, config.outputDir(), false);
```
rosiereflo, line 67: Please declare a const bool& append to make the code easier to read

sources/main.cpp
rosiereflo: yes, the MR is almost ready :)

rosiereflo (changes requested):

sources/Common/src/Log.cpp
lecourtoisflo: OK

sources/main.cpp
lecourtoisflo: OK

rosiereflo (approved):

PR #41, opened 2020-10-21
Dump log in outputDir found in config and add an clean error if the n…

rosiereflo:
…etwork file is not found

### DIFF
--- a/sources/Common/include/Log.h
+++ b/sources/Common/include/Log.h
@@ -36,8 +36,9 @@ class Log {
    * - one logger with the name of the program, containing logs from dynaflow launcher
    * - one logger named dynawo.log, containing the rest of dynawo logs
    * All loggers have the same pattern of messages and the same level, given by the input
+   * @param outputDir directory where the log should be written
    */
-  static void init(const common::Options& options);
+  static void init(const common::Options& options, const std::string& outputDir);
 };
 
 }  // namespace common
--- a/sources/Common/include/Log.h
+++ b/sources/Common/include/Log.h
@@ -34,8 +34,8 @@ class Log {
    *
    * Relies on dynawo logging system the following way:
    * - one logger with the name of the program, containing logs from dynaflow launcher
-   * - one logger named dynawo.log, containing the rest of dynawo logs
-   * All loggers have the same pattern of messages and the same level, given by the input
+   * messages patterns and  level are given by the input
+   * @param options user options
    * @param outputDir directory where the log should be written
    */
   static void init(const common::Options& options, const std::string& outputDir);
--- a/sources/Common/src/Log.cpp
+++ b/sources/Common/src/Log.cpp
@@ -15,32 +15,30 @@
 
 #include "Log.h"
 
+#include <boost/filesystem.hpp>
+
+namespace file = boost::filesystem;
+
 namespace dfl {
 namespace common {
 
 const char* Log::dynaflowLauncherLogTag = "DYNAFLOW_LAUNCHER";
 
 void
-Log::init(const common::Options& options) {
+Log::init(const common::Options& options, const std::string& outputDir) {
   using DYN::Trace;
   auto& config = options.config();
 
   std::vector<Trace::TraceAppender> appenders;
   Trace::TraceAppender appender;
+  file::path outputPath(outputDir);
+  if (!file::exists(outputPath)) {
+    file::create_directories(outputPath);
+  }
 
-  appender.setFilePath(config.programName + ".log");
+  appender.setFilePath(outputPath.append(config.programName + ".log").generic_string());
   appender.setLvlFilter(Trace::severityLevelFromString(config.dynawoLogLevel));
   appender.setTag(dynaflowLauncherLogTag);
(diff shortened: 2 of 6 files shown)


rosiereflo:
#32

sources/Common/src/Log.cpp
```diff
@@ -15,20 +15,28 @@
...
 void
-Log::init(const common::Options& options) {
+Log::init(const common::Options& options, const std::string& outputDir) {
   using DYN::Trace;
   auto& config = options.config();
 
   std::vector<Trace::TraceAppender> appenders;
   Trace::TraceAppender appender;
+  file::path outputPath(outputDir);
+  if (!file::exists(outputPath)) {
```
lecourtoisflo, line 35: maybe remove the corresponding lines in context.exportOuputs

sources/Context.h
```diff
@@ -118,7 +118,7 @@ class Context {
  private:
   ContextDef def_;                         ///< context definition
   inputs::NetworkManager networkManager_;  ///< network manager
-  inputs::Configuration config_;           ///< configuration
+  const inputs::Configuration& config_;           ///< configuration
```
lecourtoisflo, line 121: incorrect formatting

sources/main.cpp
```diff
@@ -76,17 +79,21 @@ main(int argc, char* argv[]) {
     if (!boost::filesystem::is_regular_file(dictPath)) {
       // we cannot use dictionnary errors since they are not be initialized yet
       LOG(error) << "Dictionary " << dictPath << " not found: check runtime environment" << LOG_ENDL;
+      std::cerr << "Dictionary " << dictPath << " not found: check runtime environment" << std::endl;
```
lecourtoisflo, line 82: why use cerr ? Don't we want to use the log instead ? cerr should be used only when it is unsure that log file is initialized

sources/main.cpp
```diff
@@ -76,17 +79,21 @@ main(int argc, char* argv[]) {
...
       return EXIT_FAILURE;
     }
-
-    auto& config = options.config();
-    LOG(info) << MESS(InputsInfo, config.networkFilePath, config.configPath) << LOG_ENDL;
+    if (!boost::filesystem::exists(boost::filesystem::path(runtimeConfig.networkFilePath))) {
+      LOG(error) << MESS(NetworkFileNotFound, runtimeConfig.networkFilePath) << LOG_ENDL;
+      std::cerr << MESS(NetworkFileNotFound, runtimeConfig.networkFilePath) << std::endl;
```
lecourtoisflo, line 87: same here as previous remark about cerr

sources/Context.h
```diff
@@ -53,8 +52,9 @@ class Context {
    * @brief Constructor
    *
    * @param def The context definition
+   * @param config simulation configuration
```
lecourtoisflo, line 55: it is not simulation configuration rather than DFL specific configuration

lecourtoisflo (changes requested):

sources/main.cpp
rosiereflo: It find it nicer to have a feedback in the terminal when an error happened, but that's only personal
@marcochiaramello an opinion on this?

marcochiaramello: I do agree, maybe in debug mode if possible..otherwise I'll keep it

sources/Common/src/Log.cpp
rosiereflo: done

sources/Context.h
rosiereflo: done

sources/Context.h
rosiereflo: done

sources/main.cpp
lecourtoisflo: In that case, it should be done in dynawo, in function error from Trace

rosiereflo: ok I removed them

sources/Context.h
```diff
@@ -53,8 +52,9 @@ class Context {
    * @brief Constructor
    *
    * @param def The context definition
+   * @param config configuration to use
    */
-  explicit Context(const ContextDef& def);
+  explicit Context(const ContextDef& def, const inputs::Configuration& config);
```
lecourtoisflo, line 57: explicit is no longer necessary

sources/Common/include/Log.h
```diff
@@ -36,8 +36,9 @@ class Log {
    * - one logger with the name of the program, containing logs from dynaflow launcher
    * - one logger named dynawo.log, containing the rest of dynawo logs
```
lecourtoisflo, line 37: comment not right any more. Also missing comment for parameter options from previous develpments

lecourtoisflo (changes requested):

sources/Context.h
rosiereflo: done

sources/Common/include/Log.h
rosiereflo: done

lecourtoisflo (approved):

PR #43, opened 2020-10-21
Correct diagram file link

lecourtoisflo:


sources/Outputs/include/Par.h
```diff
@@ -42,14 +42,17 @@ class Par {
      *
      * @param base basename for current simulation
      * @param filename file path for output PAR file (corresponds to basename)
+     * @param dirname the dirname of the output PAR file
```
rosiereflo, line 45: Wrong parameter name and wrong parameter order (name 'dir' below)

rosiereflo (changes requested):

sources/Context.cpp
```diff
@@ -151,15 +151,15 @@ Context::exportOutputs() {
   // create specific par
   file::path parOutput(config_.outputDir());
   parOutput.append(basename_ + ".par");
-  outputs::Par parWriter(outputs::Par::ParDefinition(basename_, parOutput.generic_string(), generators_));
+  outputs::Par parWriter(outputs::Par::ParDefinition(basename_, config_.outputDir(), parOutput.generic_string(), generators_));
```
marcochiaramello, line 154: Here I have the impression we pass twice an information, config_.outputDir() and  parOutput.generic_string().
Indeed, parOutput.genereic_string() is built just above from config_.outputDir()

marcochiaramello (changes requested):

sources/Context.cpp
lecourtoisflo: technicay correct, as parOutput = config_.outputDir() + basename_. I keep all of these to keep the spirit of putting the filename in the writer definition and to have flexibility in the future if, for one reason or another we would put the diagram file in another directory than the dyd and par. Should I restrict the information ?

sources/Outputs/include/Par.h
lecourtoisflo: OK

rosiereflo (approved):

sources/Context.cpp
rosiereflo: I don't have any preference, both point of views are OK to me.... (sorry for not helping ;))
Marco I'll let you have the final word on this.

marcochiaramello: ok, let's keep some room for flexibility :)

tests/outputs/TestPar.cpp
```diff
@@ -58,7 +59,7 @@ TEST(TestPar, write) {
                                                                      },
                                                                      4., 40., 44., 440.)};
 
-  dfl::outputs::Par parWriter(dfl::outputs::Par::ParDefinition(basename, outputDir + "/" + filename, generators));
+  dfl::outputs::Par parWriter(dfl::outputs::Par::ParDefinition(basename, dirname, outputDir + "/" + filename, generators));
```
marcochiaramello, line 62: small question: explicit separator like here "/" will work on OS platforms?

marcochiaramello (changes requested):

tests/outputs/TestPar.cpp
lecourtoisflo: I don't know but it is true that we can protect ourselves in that case. I'll do the correction. But we'll still have the issue in the config.json if we have multi-directories output dir

marcochiaramello (approved):

rosiereflo (approved):

PerninGuillaume (approved):

gautierbureau (approved):

PR #44, opened 2020-10-21
Add corrections for remote compiling on centos 7.8

lecourtoisflo:
required to run in the distant machine

### DIFF
--- a/sources/Outputs/include/Dyd.h
+++ b/sources/Outputs/include/Dyd.h
@@ -29,6 +29,31 @@
 #include <unordered_map>
 #include <vector>
 
+namespace std {
+/**
+ * @brief specialization hash for generators model types
+ *
+ * For old compilers, hashs are not implemented for unordered_map when key is an enum class.
+ * see https://en.cppreference.com/w/cpp/utility/hash for template definition
+ */
+template<>
+struct hash<dfl::algo::GeneratorDefinition::ModelType> {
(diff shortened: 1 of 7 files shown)


rosiereflo (approved):

gautierbureau:
> required to run in the distant machine

More precisely on centos 7.8

gautierbureau (approved):

sources/Outputs/include/Dyd.h
```diff
@@ -29,6 +29,31 @@
...
+ *
+ * For old compilers, hashs are not implemented for unordered_map when key is an enum class.
+ * see https://en.cppreference.com/w/cpp/utility/hash for template definition
+ */
+template<>
+struct hash<dfl::algo::GeneratorDefinition::ModelType> {
+  /// @brief Constructor
+  hash() {}
+  /**
+   * @brief Action operator
```
rosiereflo, line 44: Is it possible to avoid duplicating this code?

rosiereflo (changes requested):

sources/Outputs/include/Dyd.h
lecourtoisflo: I tried but there is a conflicts between the template we declare and the template implemented in libstd, the one that doesn't exist in the distante machine. I'm open to suggestions

rosiereflo (approved):

PR #45, opened 2020-10-22
Move install target in the right place to be able to use cmake 3.12. …

gautierbureau:
…Add compile flag for xerces compilation on centos7.8.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -98,6 +98,7 @@ find_package(Sundials REQUIRED)
 # libZIP
 find_package(libZIP 1.3.0 REQUIRED)
 # XercesC (required for libxml)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -98,6 +98,7 @@ find_package(Sundials REQUIRED)
 # libZIP
 find_package(libZIP 1.3.0 REQUIRED)
 # XercesC (required for libxml)
+set(CXX_STDFLAG "-std=c++11") # Required to test Xerces compilation
```
lecourtoisflo, line 101: Will be integrated with the same comment also in another PR

lecourtoisflo (approved):

rosiereflo (approved):

PR #48, opened 2020-10-23
Diagram validity

PerninGuillaume:
Work in progress, need to add integration test with a iidm with an invalid diagram, the tests we have at the moment have all valid diagram.


sources/Algo/src/Algo.cpp
```diff
@@ -81,16 +83,53 @@ GeneratorDefinitionAlgorithm::operator()(const NodePtr& node) {
...
+  auto it = generator.points.begin();
+  while ((allQminEqualQmax || allPEqual) && it != generator.points.end()) {
+    allQminEqualQmax = allQminEqualQmax && it->qmin == it->qmax;
+    allPEqual = allPEqual && it->p == firstP;
+    ++it;
+  }
+  valid = !allQminEqualQmax && !allPEqual;
+
+  if (!valid) {
+    LOG(info) << generator.id + "has not a valid diagram." << LOG_ENDL;
```
lecourtoisflo, line 128: This should be a warning log if we put the user to be notified or a debug log.

tests/algo/TestAlgo.cpp
```diff
@@ -259,3 +260,80 @@ TEST(Loads, base) {
     ASSERT_EQ(expected_loads[index].id, loads[index].id);
   }
 }
+
+TEST(Diagram, validDiagram) {
```
lecourtoisflo, line 264: The name of the suite should be consistent "Diagram" should be consistent with the rest of the file

sources/Algo/include/Algo.h
```diff
@@ -199,6 +199,15 @@ class GeneratorDefinitionAlgorithm : public NodeAlgorithm {
   void operator()(const NodePtr& node);
```
lecourtoisflo, line 199: Complete description of algorithm

lecourtoisflo (changes requested):
Do not commit code with WIP message

sources/Algo/include/Algo.h
PerninGuillaume: What do you mean ?

lecourtoisflo: The doxygen docuentation of the operator() provides a short description of the algorithm. It could be interesting to complete it, at least to say that some input generators can be ignored by this algorithm

sources/Algo/src/Algo.cpp
```diff
@@ -125,7 +125,7 @@ GeneratorDefinitionAlgorithm::isDiagramValid(const inputs::Generator& generator)
   valid = !allQminEqualQmax && !allPEqual;
 
   if (!valid) {
-    LOG(info) << generator.id + "has not a valid diagram." << LOG_ENDL;
+    LOG(warn) << generator.id + "has not a valid diagram." << LOG_ENDL;
```
lecourtoisflo, line 128: For error / warning / info logs,dictionnaries should be used

lecourtoisflo (changes requested):

lecourtoisflo (approved):

rosiereflo:
Please synchronize with the master to resolve the conflict

etc/DFLMessages_en_GB.dic
```diff
@@ -18,3 +18,4 @@ SimulateInfo                  =     Initializing %1% simulation
 NetworkFileNotFound           =     Network file %1% does not exist
 NodeBreakerInterface          =     Node breaker topology for voltage level interface is not supported (%1%)
 SimulationEnded               =     Simulation %1% ended sucessfully
+InvalidDiagram                =     The diagram of the generator %1% is invalid
```
rosiereflo, line 21: Please add the reason of the non-validity

sources/Algo/include/Algo.h
```diff
@@ -192,13 +192,24 @@ class GeneratorDefinitionAlgorithm : public NodeAlgorithm {
   /**
    * @brief Perform algorithm
    *
-   * Add the generators of the nodes and deducing the model to use
+   * Add the generators of the nodes and deducing the model to use. If the node is a generator with a deduced
+   * model of a diagram then the function isDiagramValid is called. If the diagram is not valid the generator
```
rosiereflo, line 196: "If the diagram is not valid the generator..." => If the diagram is not valid the default generator model is used

sources/Algo/include/Algo.h
```diff
@@ -192,13 +192,24 @@ class GeneratorDefinitionAlgorithm : public NodeAlgorithm {
   /**
    * @brief Perform algorithm
    *
-   * Add the generators of the nodes and deducing the model to use
+   * Add the generators of the nodes and deducing the model to use. If the node is a generator with a deduced
```
rosiereflo, line 195: Please rephrase: for example "validity of the generator is checked whenever it makes sense"

sources/Algo/src/Algo.cpp
```diff
@@ -81,16 +83,53 @@ GeneratorDefinitionAlgorithm::operator()(const NodePtr& node) {
   if (node_generators.size() == 1) {
     auto model = useInfiniteReactivelimits_ ? GeneratorDefinition::ModelType::SIGNALN : GeneratorDefinition::ModelType::DIAGRAM_PQ_SIGNALN;
     const auto& gen = node_generators.front();
+
+    if (model == GeneratorDefinition::ModelType::DIAGRAM_PQ_SIGNALN && !isDiagramValid(gen)) {
+      return;
```
rosiereflo, line 88: Please add a warning saying that the default model willl be used for this generator

sources/Algo/src/Algo.cpp
```diff
@@ -81,16 +83,53 @@ GeneratorDefinitionAlgorithm::operator()(const NodePtr& node) {
...
+    }
+
     generators_.emplace_back(gen.id, model, node->id, gen.points, gen.qmin, gen.qmax, gen.pmin, gen.pmax);
   } else {
     auto model =
         useInfiniteReactivelimits_ ? GeneratorDefinition::ModelType::WITH_IMPEDANCE_SIGNALN : GeneratorDefinition::ModelType::WITH_IMPEDANCE_DIAGRAM_PQ_SIGNALN;
     for (auto it = node_generators.begin(); it != node_generators.end(); ++it) {
+      if (model == GeneratorDefinition::ModelType::WITH_IMPEDANCE_DIAGRAM_PQ_SIGNALN && !isDiagramValid(*it)) {
+        continue;
```
rosiereflo, line 97: Same as above

sources/Outputs/include/Diagram.h
```diff
@@ -62,6 +62,7 @@ class Diagram {
    */
   void write();
 
+
```
rosiereflo, line 65: Please remove useless blank line

sources/Outputs/include/Diagram.h
```diff
@@ -76,8 +77,8 @@ class Diagram {
    * @param buffer The buffer to store the string that will be written to the file
    * @param table The enum determining if we write the Qmin or Qmax table
    */
+  static void writeTable(const algo::GeneratorDefinition& generator, std::stringstream& buffer, Tables table);
```
rosiereflo, line 80: why static?

sources/Algo/src/Algo.cpp
```diff
@@ -81,16 +83,53 @@ GeneratorDefinitionAlgorithm::operator()(const NodePtr& node) {
...
+      }
       generators_.emplace_back(it->id, model, node->id, it->points, it->qmin, it->qmax, it->pmin, it->pmax);
     }
   }
 }
 
+bool
+GeneratorDefinitionAlgorithm::isDiagramValid(const inputs::Generator& generator) {
+  bool valid = true;
```
rosiereflo, line 106: Please declare this at the latest moment

sources/Algo/src/Algo.cpp
```diff
@@ -81,16 +83,53 @@ GeneratorDefinitionAlgorithm::operator()(const NodePtr& node) {
...
+
+  if (generator.points.size() == 0) {
```
rosiereflo, line 108: if generator.empty

sources/Algo/src/Algo.cpp
```diff
@@ -81,16 +83,53 @@ GeneratorDefinitionAlgorithm::operator()(const NodePtr& node) {
...
+    return true;
```
rosiereflo, line 109: What does it mean to have an empty diagram?

sources/Algo/src/Algo.cpp
```diff
@@ -81,16 +83,53 @@ GeneratorDefinitionAlgorithm::operator()(const NodePtr& node) {
...
+  if (generator.points.size() == 1) {
+    LOG(warn) << MESS(InvalidDiagram, generator.id) << LOG_ENDL;
+    return false;
+  }
+
+  auto firstP = generator.points.front().p;
+  bool allQminEqualQmax = true;
+  bool allPEqual = true;
+  auto it = generator.points.begin();
+  while ((allQminEqualQmax || allPEqual) && it != generator.points.end()) {
```
rosiereflo, line 120: I always prefer for loops to while loops

sources/Algo/src/Algo.cpp
```diff
@@ -81,16 +83,53 @@ GeneratorDefinitionAlgorithm::operator()(const NodePtr& node) {
...
+    allQminEqualQmax = allQminEqualQmax && it->qmin == it->qmax;
```
rosiereflo, line 121: as soon as this is false we might break the loop and avoir useless computations

tests/algo/TestAlgo.cpp
```diff
@@ -259,3 +261,83 @@ TEST(Loads, base) {
     ASSERT_EQ(expected_loads[index].id, loads[index].id);
   }
 }
+
+TEST(Generators, validDiagram) {
+  using dfl::inputs::Generator;
+  std::vector<Generator::ReactiveCurvePoint> points({Generator::ReactiveCurvePoint(12., 44., 440.), Generator::ReactiveCurvePoint(65., 44., 440.)});
+  const Generator generator("G1", points, 3., 30., 33., 330.);
```
rosiereflo, line 268: Please test the empty and the size==1 cases

rosiereflo (changes requested):

sources/Algo/include/Algo.h
lecourtoisflo: incorrect. from algorithm point of view, if diagram is invalid, the generator is ignored

sources/Algo/src/Algo.cpp
PerninGuillaume: In log(debug) or log(warning) ?

PerninGuillaume: We already have a warning when the diagram is invalid, so should I extend the original warning instead ?

sources/Outputs/include/Diagram.h
PerninGuillaume: It does not depend on any class members, so I figured it would be more optimal

sources/Algo/src/Algo.cpp
PerninGuillaume: It means that we still have a diagram in the diagram.txt file. This diagram will be constructed from the pmin, pmax, qmin and qmax points and will give us a table(2,2)

sources/Algo/src/Algo.cpp
PerninGuillaume: I prefer the while loops, it isn't mandatory to have for loops ?

sources/Algo/src/Algo.cpp
PerninGuillaume: One Reactive Curve Point (RCP) which has qmin != qmax does not imply that the diagram is valid. This same diagram could have all its P equal for all we know, so we must continue the loop until we are sure.

sources/Algo/src/Algo.cpp
rosiereflo: the consequence of this invalidity should also be given

sources/Algo/src/Algo.cpp
rosiereflo: Could you please add a comment?

sources/Algo/src/Algo.cpp
rosiereflo: Nope, not mandatory, just more elegant

sources/Algo/include/Algo.h
rosiereflo: both are equivalent I guess

sources/Algo/src/Algo.cpp
PerninGuillaume: It boils down to personal preference

sources/Algo/src/Algo.cpp
PerninGuillaume: Of course

sources/Algo/include/Algo.h
rosiereflo: Let's keep it that way as the warning will give the consequence

sources/Algo/src/Algo.cpp
rosiereflo: absolutely :)

rosiereflo: Just to explain my preference: the main interest of for loop is to avoid having to handle the iterator inside the loop, which can be dangerous if the complexity increase. Here the algorithm is simple so it's ok to keep it that way if you prefer

marcochiaramello (approved):

rosiereflo (approved):

sources/Algo/src/Algo.cpp
```diff
@@ -81,16 +84,62 @@ GeneratorDefinitionAlgorithm::operator()(const NodePtr& node) {
...
 }
 
+bool
+GeneratorDefinitionAlgorithm::isDiagramValid(const inputs::Generator& generator) {
+  //If there are no points, the diagram will be constructed from the pmin, pmax, qmin and qmax values
+  if (generator.points.empty()) {
+    return true;
+  }
+  if (generator.points.size() == 1) {
+    LOG(warn) << MESS(InvalidDiagram, generator.id, "there was only one reactive curve point provided") << LOG_ENDL;
```
lecourtoisflo, line 112: do not use english text in dictionnary message

sources/Algo/src/Algo.cpp
```diff
@@ -81,16 +84,62 @@ GeneratorDefinitionAlgorithm::operator()(const NodePtr& node) {
...
+    if (allQminEqualQmax) {
+      errorMessage = "each reactive curve point has the same qmin and qmax";
+    }
+    if (allQminEqualQmax && allPEqual) {
+      errorMessage += " and ";
+    }
+    if (allPEqual) {
+      errorMessage += "all reactive curve points have the same p";
+    }
+    LOG(warn) << MESS(InvalidDiagram, generator.id, errorMessage) << LOG_ENDL;
```
lecourtoisflo, line 138: same remark as above. Maybe create 2 different messages

lecourtoisflo (changes requested):

lecourtoisflo (approved):

lecourtoisflo (changes requested):

sources/Algo/src/Algo.cpp
```diff
@@ -82,16 +85,61 @@ GeneratorDefinitionAlgorithm::operator()(const NodePtr& node) {
...
+  auto it = generator.points.begin();
+  while ((allQminEqualQmax || allPEqual) && it != generator.points.end()) {
+    allQminEqualQmax = allQminEqualQmax && it->qmin == it->qmax;
+    allPEqual = allPEqual && it->p == firstP;
+    ++it;
+  }
+  bool valid = !allQminEqualQmax && !allPEqual;
+
+  if (!valid) {
+    std::string errorMessage = "";
```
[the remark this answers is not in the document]
PerninGuillaume: No it did not

sources/Algo/include/Algo.h
```diff
@@ -192,13 +192,24 @@ class GeneratorDefinitionAlgorithm : public NodeAlgorithm {
   /**
    * @brief Perform algorithm
    *
-   * Add the generators of the nodes and deducing the model to use
+   * Add the generators of the nodes and deducing the model to use.
+   * Validity of the generator is checked whenever it makes sense. 
+   * If the diagram is not valid, the default generator model is used
```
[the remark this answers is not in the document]
rosiereflo: let's slit the apple in 2 like the other case: "if the diagram is not valid, the generator is ignored by the algorithm and the default dynawo behavior is used"

lecourtoisflo (approved):

gautierbureau (approved):

PR #49, opened 2020-10-27
Correct encoding for XML file exports

lecourtoisflo:
Linked to dynawo PR 1228


gautierbureau (approved):

sources/Outputs/include/Constants.h
```diff
@@ -42,6 +42,7 @@ const std::string diagramMaxTableSuffix{"_tableqmax"};              ///< Suffix
 const std::string diagramMinTableSuffix{"_tableqmin"};              ///< Suffix for the table name for qmin in diagram file
 const std::string signalNGeneratorParId{"signalNGenerator"};        ///< PAR id for generators using signal N
 const std::string impSignalNGeneratorParId{"impSignalNGenerator"};  ///< PAR id for generators using signal N with impendance
+const std::string xmlEncoding{"UTF-8"};                             ///< Default encoding for XML outputs files
```
rosiereflo, line 45: Don't we want to have this as a parameter to that all cases can be covered? for example, this may not work with special french character if I'm not mistaken

rosiereflo (changes requested):

sources/Outputs/include/Constants.h
marcochiaramello: If I remember well utf-8 is the most general, no? should be able to manage french characters too..

marcochiaramello (approved):

sources/Outputs/include/Constants.h
rosiereflo: maybe I'm not an expert, just wanted to check :)

rosiereflo (approved):

PR #52, opened 2020-10-28
Loads: add criteria for loads library choice

PerninGuillaume:

### DIFF
--- a/sources/Algo/include/Algo.h
+++ b/sources/Algo/include/Algo.h
@@ -230,7 +230,7 @@ class LoadDefinitionAlgorithm : public NodeAlgorithm {
    * @brief Constructor
    *
    * @param loads the list of loads to update
-   * @param dsoVoltageLevel the upper bound for the change of load model 
+   * @param dsoVoltageLevel Minimum voltage level of the load to be taken into account 
    */
   explicit LoadDefinitionAlgorithm(Loads& loads, double dsoVoltageLevel);
 
@@ -246,7 +246,7 @@ class LoadDefinitionAlgorithm : public NodeAlgorithm {
 
  private:
   Loads& loads_;            ///< the loads to update
-  double dsoVoltageLevel_;  ///< the upper bound to determine wether we use the default modelica model
+  double dsoVoltageLevel_;  ///< Minimum voltage level of the load to be taken into account
 };
 
 }  // namespace algo
--- a/sources/Algo/include/Algo.h
+++ b/sources/Algo/include/Algo.h
@@ -230,22 +230,23 @@ class LoadDefinitionAlgorithm : public NodeAlgorithm {
    * @brief Constructor
    *
    * @param loads the list of loads to update
+   * @param dsoVoltageLevel the upper bound for the change of load model 
    */
-  explicit LoadDefinitionAlgorithm(Loads& loads);
+  explicit LoadDefinitionAlgorithm(Loads& loads, double dsoVoltageLevel);
 
   /**
    * @brief Perform the algorithm
    *
    * Update the list with the loads of the node. Only add the loads for which the nominal voltage of the node is 
-   * superior to dsoVoltageLevel, otherwise we use the default modelica library
+   * superior to dsoVoltageLevel, otherwise the load is ignored and is handled by the default dynawo behavior
    *
    * @param node the node to process
    */
   void operator()(const NodePtr& node);
 
  private:
-  static constexpr double dsoVoltageLevel = 45.0;   ///< the upper bound to determine wether we use the default modelica model
-  Loads& loads_;  ///< the loads to update
+  Loads& loads_;            ///< the loads to update
+  double dsoVoltageLevel_;  ///< the upper bound to determine wether we use the default modelica model
 };
 
 }  // namespace algo
--- a/sources/Algo/include/Algo.h
+++ b/sources/Algo/include/Algo.h
@@ -236,13 +236,15 @@ class LoadDefinitionAlgorithm : public NodeAlgorithm {
   /**
    * @brief Perform the algorithm
    *
-   * Update the list with the loads of the node
+   * Update the list with the loads of the node. Only add the loads for which the nominal voltage of the node is 
+   * superior to dsoVoltageLevel, otherwise we use the default modelica library
    *
    * @param node the node to process
    */
   void operator()(const NodePtr& node);
 
  private:
+  static constexpr double dsoVoltageLevel = 45.0;   ///< the upper bound to determine wether we use the default modelica model
   Loads& loads_;  ///< the loads to update
 };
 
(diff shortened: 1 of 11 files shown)


sources/Algo/include/Algo.h
```diff
@@ -236,13 +236,15 @@ class LoadDefinitionAlgorithm : public NodeAlgorithm {
...
-   * Update the list with the loads of the node
+   * Update the list with the loads of the node. Only add the loads for which the nominal voltage of the node is 
+   * superior to dsoVoltageLevel, otherwise we use the default modelica library
    *
    * @param node the node to process
    */
   void operator()(const NodePtr& node);
 
  private:
+  static constexpr double dsoVoltageLevel = 45.0;   ///< the upper bound to determine wether we use the default modelica model
```
rosiereflo, line 247: Please add this as a parameter with a default value, we should not use hard-coded values

sources/Algo/src/Algo.cpp
```diff
@@ -97,6 +97,10 @@ LoadDefinitionAlgorithm::LoadDefinitionAlgorithm(Loads& loads) : NodeAlgorithm()
 
 void
 LoadDefinitionAlgorithm::operator()(const NodePtr& node) {
+  if (node->nominalVoltage < dsoVoltageLevel) {
```
rosiereflo, line 100: We should be careful with double comparisons, maybe you could use the doubleEquals method from dynawo?

rosiereflo (changes requested):

sources/Algo/include/Algo.h
```diff
@@ -236,13 +236,15 @@ class LoadDefinitionAlgorithm : public NodeAlgorithm {
   /**
    * @brief Perform the algorithm
    *
-   * Update the list with the loads of the node
+   * Update the list with the loads of the node. Only add the loads for which the nominal voltage of the node is 
+   * superior to dsoVoltageLevel, otherwise we use the default modelica library
```
lecourtoisflo, line 240: For me it is not relevant to say that we use the default modelica model. This is the dynawo behaviour. The comments should only state that load is ignored if criteria is not passed

lecourtoisflo (changes requested):

sources/Algo/include/Algo.h
rosiereflo: moreover it is not a modelica library, we can split the apple in 2 and says that "otherwise the load is ignored and is handled by the default dynawo behavior"

sources/Algo/src/Algo.cpp
PerninGuillaume: How could I use this doubleEquals function when I am testing a < b ?

rosiereflo: if (doubleNotEquals(node->nominalVoltage, dsoVoltageLevel && node->nominalVoltage < dsoVoltageLevel)

PerninGuillaume: Ok

sources/Algo/src/Algo.cpp
```diff
@@ -17,6 +17,7 @@
 
 #include "Algo.h"
 
+#include "DYNCommon.h"
```
lecourtoisflo, line 20: formatting not done properly. Also, since it is an exterior header, you should use <>

sources/Algo/include/Algo.h
```diff
@@ -230,22 +230,23 @@ class LoadDefinitionAlgorithm : public NodeAlgorithm {
...
    *
    * @param node the node to process
    */
   void operator()(const NodePtr& node);
 
  private:
-  static constexpr double dsoVoltageLevel = 45.0;   ///< the upper bound to determine wether we use the default modelica model
-  Loads& loads_;  ///< the loads to update
+  Loads& loads_;            ///< the loads to update
+  double dsoVoltageLevel_;  ///< the upper bound to determine wether we use the default modelica model
```
rosiereflo, line 249: Please update the comment

tests/inputs/TestConfig.cpp
```diff
@@ -22,6 +22,7 @@ TEST(Config, Nominal) {
   ASSERT_TRUE(config.useVSCAsGenerators());
   ASSERT_TRUE(config.useLCCAsLoads());
   ASSERT_EQ("/tmp", config.outputDir());
+  ASSERT_EQ(45.0, config.getDsoVoltageLevel());
```
lecourtoisflo, line 25: The purpose of the nominal test is to check that a value different from the default is taken into account

tests/inputs/res/config.json
```diff
@@ -7,6 +7,7 @@
     "AutomaticSlackBusOn": "false",
     "VSCAsGenerators": "true",
     "LCCAsLoads": "true",
-    "OutputDir": "/tmp"
+    "OutputDir": "/tmp",
+    "DsoVoltageLevel": 45.0
```
lecourtoisflo, line 11: Use a value different from the default value

sources/Algo/include/Algo.h
```diff
@@ -230,22 +230,23 @@ class LoadDefinitionAlgorithm : public NodeAlgorithm {
    * @brief Constructor
    *
    * @param loads the list of loads to update
+   * @param dsoVoltageLevel the upper bound for the change of load model
```
rosiereflo, line 233: This is a lower bound. I would propose "Minimum voltage level of the load to be taken into account".

lecourtoisflo (changes requested):

sources/Inputs/include/Configuration.h
```diff
@@ -107,6 +107,15 @@ class Configuration {
     return outputDir_;
   }
 
+  /**
+   * @brief Retrieves the upper bound for the change of load model
```
rosiereflo, line 111: please update the comment

sources/Inputs/include/Configuration.h
```diff
@@ -116,6 +125,7 @@ class Configuration {
   bool useVSCAsGenerators_ = false;                                             ///< VSC considered as generators
   bool useLCCAsLoads_ = false;                                                  ///< LCC are considered as generators
   std::string outputDir_ = boost::filesystem::current_path().generic_string();  ///< Directory for output files
+  double dsoVoltageLevel_ = 45.0;                                               ///< the upper bound to determine wether we use the default modelica model
```
rosiereflo, line 128: please update the comment

tests/inputs/res/config.json
rosiereflo, line 11: Please put another value to test that the value is properly read

rosiereflo (changes requested):

sources/Algo/src/Algo.cpp
PerninGuillaume: What do you meant about formatting ?

lecourtoisflo: formatting by clang-format

rosiereflo (approved):

marcochiaramello (approved):

lecourtoisflo (approved):

PR #59, opened 2020-11-04
Multiple-diagrams: create a diagram file for each generator 

PerninGuillaume:
Before merging this branch, wide tests should be conducted to see if the division of diagram files is indeed faster


sources/Outputs/src/Diagram.cpp
```diff
@@ -28,30 +29,27 @@ Diagram::Diagram(DiagramDefinition&& def) : def_{std::forward<DiagramDefinition>
...
+
     writeTable(generator, buffer, Tables::TABLE_QMIN);
     writeTable(generator, buffer, Tables::TABLE_QMAX);
-  }
-
-  if (!empty) {
-    //Put buffer into file
-    std::ofstream ofs(def_.filename, std::ofstream::out);
+    boost::filesystem::path dir(def_.directoryPath);
+    std::string filename = dir.append(generator.id + outputs::constants::diagramFileSuffixExt).generic_string();
```
lecourtoisflo, line 51: Use a function in constants to generate the filename of diagrams files

lecourtoisflo (changes requested):

tests/outputs/TestDiagram.cpp
```diff
@@ -8,22 +8,43 @@
...
 //
 
+#include "Constants.h"
 #include "Diagram.h"
 #include "Tests.h"
 
 #include <boost/filesystem.hpp>
 
+static void
+testMultiplesFilesEquality(const std::vector<dfl::algo::GeneratorDefinition>& generators, boost::filesystem::path outputDirectory, std::string basename,
```
rosiereflo, line 18: const & for everything

lecourtoisflo (approved):

tests/outputs/TestDiagram.cpp
```diff
@@ -114,43 +130,37 @@ TEST(Diagram, writeWithCurveAndDefaultPoints) {
...
+  std::string prefixDir = "Empty";
+  boost::filesystem::path outputDirectory("results");
+  outputDirectory.append(basename);
 
-  if (!boost::filesystem::exists(outputPath)) {
-    boost::filesystem::create_directories(outputPath);
+  if (!boost::filesystem::exists(outputDirectory)) {
+    boost::filesystem::create_directories(outputDirectory);
   }
+  outputDirectory.append(prefixDir + dfl::outputs::constants::diagramDirectorySuffix);
```
rosiereflo, line 150: Please make sure the directory is empty

rosiereflo (changes requested):

tests/outputs/TestDiagram.cpp
PerninGuillaume: Do you mean I have to test twice the fact that it is empty ? It is already done 
` std::ifstream stream(directoryPath);
    ASSERT_TRUE(stream.fail()) << "The Diagram directory " << directoryPath << " has been created to even though there are no generators which model requires it."
                               << std::endl; ` 
here

sources/Outputs/src/Diagram.cpp
```diff
@@ -28,30 +29,27 @@ Diagram::Diagram(DiagramDefinition&& def) : def_{std::forward<DiagramDefinition>
...
+    std::ofstream ofs(filename, std::ofstream::out);
```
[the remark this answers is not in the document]
PerninGuillaume: Did you mean by "the diagram is empty" the case where it has no Reactive curve points ?

tests/outputs/TestDiagram.cpp
rosiereflo: ok!

sources/Outputs/src/Diagram.cpp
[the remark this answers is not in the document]
rosiereflo: yes, but I'm not sure of this one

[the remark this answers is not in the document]
PerninGuillaume: When we have no Reactive curve points in a diagram we dump it with the default values of pmin,pmax,qmin and qmax

rosiereflo (approved):

marcochiaramello (approved):

PR #60, opened 2020-11-09
Par-fields: added Pref and Uref fields

PerninGuillaume:

Added Pref and Uref fields in par file, but need to know if they should be added to the writeGenerator function too.

Changed the reference of Load_P0Pu and load_Q0Pu

### DIFF
--- a/sources/Outputs/src/Par.cpp
+++ b/sources/Outputs/src/Par.cpp
@@ -166,6 +166,8 @@ Par::writeGenerator(const algo::GeneratorDefinition& def, const std::string& bas
   set->addReference(helper::buildReference("generator_Q0Pu", "q_pu", "DOUBLE"));
   set->addReference(helper::buildReference("generator_U0Pu", "v_pu", "DOUBLE"));
   set->addReference(helper::buildReference("generator_UPhase0", "angle_pu", "DOUBLE"));
+  set->addReference(helper::buildReference("generator_Pref0Pu", "targetP_pu", "DOUBLE"));
+  set->addReference(helper::buildReference("generator_Uref0Pu", "targetV_pu", "DOUBLE"));
 
   return set;
 }
--- a/sources/Outputs/src/Par.cpp
+++ b/sources/Outputs/src/Par.cpp
@@ -86,6 +86,8 @@ Par::updateSignalNGenerator(boost::shared_ptr<parameters::ParametersSet> set) {
   set->addReference(helper::buildReference("generator_Q0Pu", "q_pu", "DOUBLE"));
   set->addReference(helper::buildReference("generator_U0Pu", "v_pu", "DOUBLE"));
   set->addReference(helper::buildReference("generator_UPhase0", "angle_pu", "DOUBLE"));
+  set->addReference(helper::buildReference("generator_Pref0Pu", "targetP_pu", "DOUBLE"));
+  set->addReference(helper::buildReference("generator_Uref0Pu", "targetV_pu", "DOUBLE"));
 }
 
 std::vector<boost::shared_ptr<parameters::ParametersSet>>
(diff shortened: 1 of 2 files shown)


cossartque (approved):

sources/Outputs/src/Par.cpp
```diff
@@ -86,6 +86,8 @@ Par::updateSignalNGenerator(boost::shared_ptr<parameters::ParametersSet> set) {
   set->addReference(helper::buildReference("generator_Q0Pu", "q_pu", "DOUBLE"));
   set->addReference(helper::buildReference("generator_U0Pu", "v_pu", "DOUBLE"));
   set->addReference(helper::buildReference("generator_UPhase0", "angle_pu", "DOUBLE"));
+  set->addReference(helper::buildReference("generator_Pref0Pu", "targetP_pu", "DOUBLE"));
+  set->addReference(helper::buildReference("generator_Uref0Pu", "targetV_pu", "DOUBLE"));
```
marcochiaramello, line 90: Pref0Pu and Uref0Pu should be added also in writeGenerator function

marcochiaramello (changes requested):

sources/Outputs/src/Par.cpp
```diff
@@ -86,6 +86,8 @@ Par::updateSignalNGenerator(boost::shared_ptr<parameters::ParametersSet> set) {
   set->addReference(helper::buildReference("generator_Q0Pu", "q_pu", "DOUBLE"));
   set->addReference(helper::buildReference("generator_U0Pu", "v_pu", "DOUBLE"));
   set->addReference(helper::buildReference("generator_UPhase0", "angle_pu", "DOUBLE"));
+  set->addReference(helper::buildReference("generator_Pref0Pu", "targetP_pu", "DOUBLE"));
```
lecourtoisflo, line 89: Does it have an impact on results of the simulation (MAIN tests) ? If not add a main test that is relevant for these changes

lecourtoisflo (changes requested):

sources/Outputs/src/Par.cpp
PerninGuillaume: In the result TestPar.par the line
`<reference type="DOUBLE" name="generator_Pref0Pu" origData="IIDM" origName="targetP_pu"/>`
was added, is it enough ?

marcochiaramello: Uref0Pu should be there too, I think

PerninGuillaume: Yes it is also here
`<reference type="DOUBLE" name="generator_Uref0Pu" origData="IIDM" origName="targetV_pu"/>`

lecourtoisflo: I was talking about the IIDM output for the whole simulation, performed in MAIN tests. I don't know exactly whet is the rational for the modification but if it corresponds to a functionnality, it should be checked in unit test and an TNR (MAIN tests) should cover it.

PerninGuillaume: @marcochiaramello Does this PR introduces change that are worth to be tested in an integration test ? That's what florent meant in these comments.
If so, could you provide an iidm that could be used for this test ?

sources/Outputs/src/Par.cpp
cossartque: Same remark as Marco

cossartque (approved):

sources/Outputs/src/Par.cpp
PerninGuillaume: This has been corrected in a recent commit

cossartque: I saw it just after my comment, my bad!

PerninGuillaume: No problem

rosiereflo (approved):

marcochiaramello (approved):

sources/Outputs/src/Par.cpp
rosiereflo: Was this point solved?

marcochiaramello: In my memories yes, @PerninGuillaume do you confirm?

PerninGuillaume: Yes

lecourtoisflo (approved):

PR #62, opened 2020-11-16
ActivePowerCompensation: added option

PerninGuillaume:
### For the integration tests:
 As the integration tests **launch** and **special_characters** were affected by this PR, I have modified config_special_characters.json to add the line "ActivePowerCompensation": "P". Before the PR the default behavior was having the ActivePowerCompensation equal to P, thus, the outputIIDM of the test special_characters does not need to be changed.

For the test config_launch I have not added the line "ActivePowerCompensation": "P", thus I had to change the outputIIDM to make the test pass. However, to not erase what the launch test previously tested, I have added another test  **launch_P**  that has the same outputIIDM as had the test launch.


sources/Inputs/include/Configuration.h
```diff
@@ -116,16 +116,31 @@ class Configuration {
...
+  bool useInfiniteReactiveLimits_ = false;                                          ///< infinite reactive limits
+  bool isPSTRegulationOn_ = true;                                                   ///< PST regulation on
+  bool isSVCRegulationOn_ = true;                                                   ///< SVC regulation on
+  bool isShuntRegulationOn_ = true;                                                 ///< Shunt regulation on
+  bool isAutomaticSlackBusOn_ = true;                                               ///< automatic slack bus on
+  bool useVSCAsGenerators_ = false;                                                 ///< VSC considered as generators
+  bool useLCCAsLoads_ = false;                                                      ///< LCC are considered as generators
+  std::string outputDir_ = boost::filesystem::current_path().generic_string();      ///< Directory for output files
+  double dsoVoltageLevel_ = 45.0;                                                   ///< Minimum voltage level of the load to be taken into account
+  ActivePowerCompensation activePowerCompensation = ActivePowerCompensation::PMAX;  ///< Type of active power compensation
```
lecourtoisflo, line 143: private members have the suffix '_'

sources/Inputs/include/Configuration.h
```diff
@@ -116,16 +116,31 @@ class Configuration {
     return dsoVoltageLevel_;
   }
 
+  /**
+   * @brief type of active power compensation for generator
+   */
+  enum class ActivePowerCompensation { P, TARGET_P, PMAX };
```
lecourtoisflo, line 122: Each possible value should be commented

sources/Inputs/src/Configuration.cpp
```diff
@@ -41,6 +42,28 @@ updateValue(T& value, const boost::property_tree::ptree& tree, const std::string
...
+  std::map<std::string, Configuration::ActivePowerCompensation> enumResolver = {{"P", Configuration::ActivePowerCompensation::P},
+                                                                                {"targetP", Configuration::ActivePowerCompensation::TARGET_P},
+                                                                                {"PMax", Configuration::ActivePowerCompensation::PMAX}};
+  std::string apcString;
+  helper::updateValue(apcString, tree, "ActivePowerCompensation");
+  auto it = enumResolver.find(apcString);
+  if (it != enumResolver.end()) {
+    activePowerCompensation = it->second;
+  } else if (!apcString.empty()) {
+    //LOG(warn) << MESS(BadActivePowerCompensation) << LOG_ENDL; //This doesn't work, I think it always expect args
```
lecourtoisflo, line 63: It should be this way instead of raw string comment for warn logs. I guess that MESS macro doesn't support having no argument so this is why you uses raw string. Maybe use a message like:
"ActivePowerCompensation was given with the unsupported value %1%. Supported values are: PMAX, P or targetP. Default value PMAX will be used"

sources/Outputs/include/Par.h
```diff
@@ -102,7 +108,7 @@ class Par {
    *
    * @returns nullptr if the generator use a SignalN model, the parameter set if not
    */
-  static boost::shared_ptr<parameters::ParametersSet> writeGenerator(const algo::GeneratorDefinition& def, const std::string& basename,
+  boost::shared_ptr<parameters::ParametersSet> writeGenerator(const algo::GeneratorDefinition& def, const std::string& basename,
```
lecourtoisflo, line 111: Keep the methods static and put the active power compensation in argument. The rational behind the 'static' pattern was to only use the required member to build the part of the file, allowing to better understand which part is using which data.

sources/Outputs/src/Par.cpp
```diff
@@ -157,7 +169,18 @@ Par::writeGenerator(const algo::GeneratorDefinition& def, const std::string& bas
...
 
-  set->addReference(helper::buildReference("generator_PNom", "p_pu", "DOUBLE"));
+  switch (def_.activePowerCompensation) {
+  case dfl::inputs::Configuration::ActivePowerCompensation::P:
+    set->addReference(helper::buildReference("generator_PNom", "p_pu", "DOUBLE"));
+    break;
+  case dfl::inputs::Configuration::ActivePowerCompensation::TARGET_P:
+    set->addReference(helper::buildReference("generator_PNom", "targetP_pu", "DOUBLE"));
+    break;
+  case dfl::inputs::Configuration::ActivePowerCompensation::PMAX:
```
lecourtoisflo, line 179: Add a comment or do nothing case for default, to explain that ther is no default case by definition of the enum.

sources/Outputs/src/Par.cpp
```diff
@@ -81,7 +81,19 @@ Par::updateSignalNGenerator(boost::shared_ptr<parameters::ParametersSet> set) {
...
-  set->addReference(helper::buildReference("generator_PNom", "p_pu", "DOUBLE"));
+
+  switch (def_.activePowerCompensation) {
+  case dfl::inputs::Configuration::ActivePowerCompensation::P:
+    set->addReference(helper::buildReference("generator_PNom", "p_pu", "DOUBLE"));
+    break;
+  case dfl::inputs::Configuration::ActivePowerCompensation::TARGET_P:
+    set->addReference(helper::buildReference("generator_PNom", "targetP_pu", "DOUBLE"));
+    break;
+  case dfl::inputs::Configuration::ActivePowerCompensation::PMAX:
```
lecourtoisflo, line 92: Add a comment or do nothing case for default, to explain that ther is no default case by definition of the enum.

sources/Inputs/include/Configuration.h
lecourtoisflo, line 143: Was this default value discussed ? It seems that the code before behaves with P instead of PMAX.

lecourtoisflo (changes requested):

sources/Inputs/src/Configuration.cpp
```diff
@@ -41,6 +42,28 @@ updateValue(T& value, const boost::property_tree::ptree& tree, const std::string
     value = value_opt->get_value<T>();
   }
 }
+
+/**
+ * @brief Helper function to update the active power compensation parameter
```
rosiereflo, line 47: Please specify here that if the value is incorrect then the value is not updated

sources/Inputs/src/Configuration.cpp
```diff
@@ -41,6 +42,28 @@ updateValue(T& value, const boost::property_tree::ptree& tree, const std::string
...
+                                                                                {"targetP", Configuration::ActivePowerCompensation::TARGET_P},
+                                                                                {"PMax", Configuration::ActivePowerCompensation::PMAX}};
+  std::string apcString;
+  helper::updateValue(apcString, tree, "ActivePowerCompensation");
+  auto it = enumResolver.find(apcString);
+  if (it != enumResolver.end()) {
+    activePowerCompensation = it->second;
+  } else if (!apcString.empty()) {
+    //LOG(warn) << MESS(BadActivePowerCompensation) << LOG_ENDL; //This doesn't work, I think it always expect args
+    LOG(warn) << "The ActivePowerCompensation parameter was wrongly entered in the config file, the different possibility are : PMAX, P or targetP. The default value of PMAX will be used." << LOG_ENDL;
```
rosiereflo, line 64: What if this option is not in the config and that we want to use the default value?

sources/Inputs/include/Configuration.h
PerninGuillaume: @marcochiaramello Any ideas on what the comments would look like for P, TARGET_P and PMAX ?

rosiereflo (changes requested):

sources/Inputs/src/Configuration.cpp
PerninGuillaume: This is what the test !apcString.empty() check. If the option is not in the config the string won't be updated by helper::updateValue.
It may not be very clear, if you want me to change it or add a comment, please say so.

lecourtoisflo: helper::updateValue does not update the value if it doesn't exist, resulting in an empty string if the config is not given. Since the default configuration is set as a default value for the member, it will be OK.

sources/Outputs/src/Par.cpp
PerninGuillaume: Do you want me to factorize the case ?

lecourtoisflo: No, juste the info for the default case. It is a good practice to help maintenability and automatic verification by tools

sources/Inputs/include/Configuration.h
PerninGuillaume: The default value has been agreed upon before on mail

lecourtoisflo (changes requested):
TODO to do

tests/algo/TestAlgo.cpp
```diff
@@ -254,7 +252,9 @@ TEST(Loads, base) {
   nodes[4]->loads.emplace_back("05");
 
   dfl::algo::LoadDefinitionAlgorithm::Loads loads;
-  dfl::algo::LoadDefinitionAlgorithm algo(loads, config.getDsoVoltageLevel());
+  //TODO remove config.json
```
lecourtoisflo, line 255: Comment to remove

lecourtoisflo (changes requested):

sources/Inputs/src/Configuration.cpp
```diff
@@ -41,6 +42,28 @@ updateValue(T& value, const boost::property_tree::ptree& tree, const std::string
     value = value_opt->get_value<T>();
   }
 }
+
+/**
+ * @brief Helper function to update the active power compensation parameter, if the value of ActivePowerCompensation
+ * is not one of the three available, then the default value is used
```
rosiereflo, line 48: Please specify that activePowerCompensation is not set when the parameter is not found (the default value used only because it as set to this value before this method).

rosiereflo (changes requested):

sources/Inputs/include/Configuration.h
marcochiaramello: Something like:
P -> active power mismatch compensation proportional to active power injection P
TARGET_P -> active power mismatch compensation proportional to targetP
PMAX -> active power mismatch compensation proportional to generator maximal active power PMax

@cossartque any opinion on this?
Thanks

marcochiaramello: let's go for this then

sources/Inputs/src/Configuration.cpp
PerninGuillaume: Should I also put in the comment that the default value is used because it was set beforehand ?

rosiereflo (approved):

marcochiaramello (approved):

lecourtoisflo (approved):

cossartque (approved):

PR #65, opened 2020-11-16
Properly catch dynawo simulation errors in log

rosiereflo:


sources/main.cpp
```diff
@@ -48,13 +48,14 @@ initializeDynawo(const std::string& locale) {
 
 int
 main(int argc, char* argv[]) {
-  try {
     boost::timer timerGlobal;
```
lecourtoisflo, line 51: alinea incorrect

sources/main.cpp
```diff
@@ -48,13 +48,14 @@ initializeDynawo(const std::string& locale) {
...
     boost::timer timerInit;
     DYN::Trace::init();
     dfl::common::Options options;
 
     auto parsing_status = options.parse(argc, argv);
```
lecourtoisflo, line 56: don't we want to catch exceptions on parsing ?

sources/main.cpp
```diff
@@ -115,32 +117,36 @@ main(int argc, char* argv[]) {
...
-    // Traces must be re-initiliazed to append to current DynaflowLauncher log file as simulation had modified it
+    // Traces must be re-initialized to append to current DynaflowLauncher log file as simulation had modified it
     dfl::common::Log::init(options, config.outputDir(), continuePreviousLogFile);
     LOG(info) << MESS(SimulationEnded, context.basename(), timerSimu.elapsed()) << LOG_ENDL;
     LOG(info) << " ============================================================ " << LOG_ENDL;
     LOG(info) << MESS(DFLEnded, context.basename(), timerGlobal.elapsed()) << LOG_ENDL;
 
     return EXIT_SUCCESS;
   } catch (DYN::Error& e) {
+    dfl::common::Log::init(options, outputDir, true);
```
lecourtoisflo, line 128: We cannot garrante that an exception doesn't occur here

lecourtoisflo (changes requested):
Maybe use a boolean to determine if we must reset the log or not according to exception origin

sources/main.cpp
rosiereflo: yes, but without this line the standard error in dynawo will never appear in the log, so I though it was a correct tradeoff?

PR #67, opened 2020-11-17
check diagram consistency also when reactive capability curve is not defined

marcochiaramello:


rosiereflo (approved):

sources/Algo/src/Algo.cpp
```diff
@@ -105,10 +105,21 @@ GeneratorDefinitionAlgorithm::operator()(const NodePtr& node) {
 
 bool
 GeneratorDefinitionAlgorithm::isDiagramValid(const inputs::Generator& generator) {
-  //If there are no points, the diagram will be constructed from the pmin, pmax, qmin and qmax values
+  //If there are no points, the diagram will be constructed from the pmin, pmax, qmin and qmax values.
+  //We check the validity of pmin,pmax and qmin,qmax values
   if (generator.points.empty()) {
+    if (generator.pmin == generator.pmax) {
```
lecourtoisflo, line 111: don't perform equality test on double: use DYN::doubleEquals

lecourtoisflo (changes requested):

sources/Algo/src/Algo.cpp
marcochiaramello: done, thanks

lecourtoisflo (approved):

PR #68, opened 2020-11-17
Correct messages

### LINKED ISSUE #66: Logging message
The message system should work the same way as in dynawo:
- no key => default message
- key present but message not compatible => exception

lecourtoisflo:
Fixes #66 

We removed the dico configuration in algo test in order to not confuse a future developer with it and focus only on testing algo library

### DIFF
--- a/sources/Common/include/Message.hpp
+++ b/sources/Common/include/Message.hpp
@@ -19,6 +19,7 @@
 #include "DicoKeys.h"
 
 #include <boost/format.hpp>
+#include <sstream>
 #include <vector>
 
 namespace dfl {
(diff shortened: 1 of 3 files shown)


rosiereflo (approved):

PerninGuillaume (commented):
Should not we have a test when the dico is not loaded to test that the default message is indeed loaded ?

lecourtoisflo:
> Should not we have a test when the dico is not loaded to test that the default message is indeed loaded ?

Not possible by design of the test. Because dictionnary is a singleton, if we don't load a dictionnary before use, we will not be able to update it after, leading to all tests to fail.

PerninGuillaume (approved):

PR #72, opened 2020-11-18
Use persistant trace functionality from dynawo to properly catch all …

rosiereflo:
…traces

### DIFF
--- a/sources/main.cpp
+++ b/sources/main.cpp
@@ -116,35 +116,43 @@ main(int argc, char* argv[]) {
     context.execute();
 
     // Traces must be re-initiliazed to append to current DynaflowLauncher log file as simulation had modified it
-    dfl::common::Log::init(options, config.outputDir(), continuePreviousLogFile);
     LOG(info) << MESS(SimulationEnded, context.basename(), timerSimu.elapsed()) << LOG_ENDL;
     LOG(info) << " ============================================================ " << LOG_ENDL;
     LOG(info) << MESS(DFLEnded, context.basename(), timerGlobal.elapsed()) << LOG_ENDL;
+    DYN::Trace::resetPersistantCustomAppenders();
 
     return EXIT_SUCCESS;
   } catch (DYN::Error& e) {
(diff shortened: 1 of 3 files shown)


sources/main.cpp
```diff
@@ -116,35 +116,43 @@ main(int argc, char* argv[]) {
     context.execute();
 
     // Traces must be re-initiliazed to append to current DynaflowLauncher log file as simulation had modified it
-    dfl::common::Log::init(options, config.outputDir(), continuePreviousLogFile);
     LOG(info) << MESS(SimulationEnded, context.basename(), timerSimu.elapsed()) << LOG_ENDL;
     LOG(info) << " ============================================================ " << LOG_ENDL;
     LOG(info) << MESS(DFLEnded, context.basename(), timerGlobal.elapsed()) << LOG_ENDL;
+    DYN::Trace::resetPersistantCustomAppenders();
```
lecourtoisflo, line 122: Why perform a reset here since the program ends ?

sources/main.cpp
```diff
@@ -116,35 +116,43 @@ main(int argc, char* argv[]) {
     context.execute();
 
     // Traces must be re-initiliazed to append to current DynaflowLauncher log file as simulation had modified it
-    dfl::common::Log::init(options, config.outputDir(), continuePreviousLogFile);
```
lecourtoisflo, line 119: Maybe modify Log::init function since we no longer need the continuePreviousLogFile argument

lecourtoisflo (changes requested):

marcochiaramello (approved):

sources/main.cpp
rosiereflo: solved

sources/main.cpp
rosiereflo: solved

lecourtoisflo (approved):

PerninGuillaume (approved):

gautierbureau (approved):

PR #73, opened 2020-11-19
Correct minors and styling

lecourtoisflo:


tests/outputs/TestDiagram.cpp
```diff
@@ -159,8 +155,8 @@ TEST(Diagram, writeEmpty) {
   std::string directoryPath = outputDirectory.generic_string();
   try {
     std::ifstream stream(directoryPath);
-    ASSERT_TRUE(stream.fail()) << "The Diagram directory " << directoryPath << " has been created to even though there are no generators which model requires it."
-                               << std::endl;
+    ASSERT_TRUE(stream.fail()) << "The Diagram directory " << directoryPath
+                               << " has been created to even though there are no generators which model requires it." << std::endl;
```
rosiereflo, line 159: Let's correct this as we modify this part of the code: even though it is not used by any generator model

rosiereflo (changes requested):

tests/outputs/TestDiagram.cpp
lecourtoisflo: OK

rosiereflo (approved):

PerninGuillaume (approved):

Issue #76, opened 2020-11-23
Precision values in unit tests


lecourtoisflo:
Values in MAIN tests, depending on the tester, have a precision of 10^-12, which is acceptable. Reference should be tested accordingly.

PR #78, opened 2020-11-23
Improve debug testing

lecourtoisflo:
Also update tests references, as majority has the same results (see #76 )

### DIFF
--- a/sources/Outputs/src/Job.cpp
+++ b/sources/Outputs/src/Job.cpp
@@ -41,6 +41,8 @@
 #include <JOBSolverEntryFactory.h>
 #include <boost/make_shared.hpp>
 #include <fstream>
+#include <xml/sax/formatter/AttributeList.h>
+#include <xml/sax/formatter/Formatter.h>
 
 namespace dfl {
 namespace outputs {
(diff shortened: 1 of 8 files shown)


sources/Outputs/src/Job.cpp
```diff
@@ -130,5 +132,106 @@ Job::writeOutputs() {
...
+  attrs.clear();
+  formatter->endElement();  // simulation
+
+  // outputs
+  auto outputs = jobEntry->getOutputsEntry();
+  attrs.add("directory", outputs->getOutputsDirectory());
+  formatter->startElement("dyn", "outputs", attrs);
+  attrs.clear();
+
+  auto logs = outputs->getLogsEntry();
```
rosiereflo, line 214: I think the line to generate the final IIDM is missing no?

rosiereflo (changes requested):

sources/Outputs/src/Job.cpp
lecourtoisflo: No, it is done automattically by the formatter

rosiereflo (approved):

PerninGuillaume (commented):
Perhaps this exportJob function should be tested

rosiereflo:
In my humble opinion it is not critical to test a debug method

PerninGuillaume (approved):

PR #81, opened 2020-11-25
HvdcLine: added the hvdc line model

PerninGuillaume:
This PR adds the hvdc line model.

In order for this PR to work it needs the line `DYNConverterInterface.h` that I added in the `DATA_INTERFACE_INCLUDE_HEADERS` in the CmakeList.txt of dynawo in sources/Modeler/DataInterface.


sources/Algo/include/Algo.h
```diff
@@ -259,5 +260,74 @@ class LoadDefinitionAlgorithm : public NodeAlgorithm {
...
 
+/**
+ * @brief Hvdc line definition for algorithms
+ */
+struct HvdcLineDefinition {
+  using HvdcLines = std::vector<HvdcLineDefinition>;  ///< alias for list of hvdcLines
+
+  using HvdcLineId = std::string;  ///< HvdcLine id definition
+
+  /** @brief enum to determine how the converters of this hvdc line are connected inside the network
```
lecourtoisflo, line 271: Avoid a brief that long. Cut in half between a brief and a detailed

sources/Algo/include/Algo.h
```diff
@@ -259,5 +260,74 @@ class LoadDefinitionAlgorithm : public NodeAlgorithm {
...
+                              Position position) :
+      id{id},
+      converterType{converterType},
+      converter1{converter1},
+      converter2{converter2},
+      position{position} {}
+
+  const HvdcLineId id;                                       ///< HvdcLine id
+  const inputs::HvdcLine::ConverterType converterType;       ///< type of converter of the hvdc line
+  inputs::ConverterInterface converter1;  ///< TODO do we need the pointer, first converter
```
rosiereflo, line 300: Would be better to use a reference to avoid copy

sources/Algo/include/Algo.h
```diff
@@ -259,5 +260,74 @@ class LoadDefinitionAlgorithm : public NodeAlgorithm {
...
+   * @param hvdcLines the list of hvdc lines to update
+   */
+  explicit ControllerInterfaceDefinitionAlgorithm(std::vector<HvdcLineDefinition>& hvdcLines);
+
+  /**
+   * @brief Perform the algorithm
+   *
+   * Update the list with the hvdc line of the converters of the node. 
+   * This function also put the position of the converters compared to the main connex component in the newly created hvdc line definition.
+   * Indeed, the node that are processed through this operator() are only those that are in the main connex component, which let us deduce the position of the
```
rosiereflo, line 322: This last sentence is a pre-condition of this algorithm, so I would prefer something like : pre-condition: the nodes used as parameter of this operator should be nodes of the main connex component only

sources/Algo/include/Algo.h
```diff
@@ -259,5 +260,74 @@ class LoadDefinitionAlgorithm : public NodeAlgorithm {
...
+   * @brief Constructor
+   *
+   * @param id the HvdcLine id
+   * @param converterType type of converter of the hvdc line
+   * @param converter1 first converter connected to the hvdc line
+   * @param converter2 second converter connected to the hvdc line
+   * @param position position of the converters of this hvdc line compared to the main connex component
+   */
+  explicit HvdcLineDefinition(const HvdcLineId& id, const inputs::HvdcLine::ConverterType converterType,
+                              inputs::ConverterInterface converter1, inputs::ConverterInterface converter2,
```
lecourtoisflo, line 290: converter interfaces are struct so use const& to copy

sources/Algo/include/Algo.h
lecourtoisflo, line 300: TODO ?

sources/Algo/src/Algo.cpp
```diff
@@ -162,6 +162,25 @@ LoadDefinitionAlgorithm::operator()(const NodePtr& node) {
...
+
+void
+ControllerInterfaceDefinitionAlgorithm::operator()(const NodePtr& node) {
+  for (auto converter : node->converterInterfaces) {
+    auto hvdcLine = converter.hvdcLine;
+    HvdcLineDefinition::Position position;
+    if (converter.converterId == hvdcLine->converter1.converterId) {
+      position = HvdcLineDefinition::Position::FIRST_IN_MAIN_COMPONENT;
+    } else if (converter.converterId == hvdcLine->converter2.converterId) {
+      position = HvdcLineDefinition::Position::SECOND_IN_MAIN_COMPONENT;
```
rosiereflo, line 177: So if I understand well if both are in the main connex component we fail today?

sources/Algo/src/Algo.cpp
```diff
@@ -162,6 +162,25 @@ LoadDefinitionAlgorithm::operator()(const NodePtr& node) {
...
+    } else {
+      throw("The hvdcLine was badly initialized.");
```
lecourtoisflo, line 179: Exception must be caught using the dictionnary if caught globaly

sources/Context.cpp
```diff
@@ -135,8 +135,7 @@ Context::exportOutputs() {
   // create specific par
   file::path parOutput(config_.outputDir());
   parOutput.append(basename_ + ".par");
-  outputs::Par parWriter(
-      outputs::Par::ParDefinition(basename_, config_.outputDir(), parOutput.generic_string(), generators_, config_.getActivePowerCompensation()));
+  outputs::Par parWriter(outputs::Par::ParDefinition(basename_, config_.outputDir(), parOutput.generic_string(), generators_, hvdcLines_, config_.getActivePowerCompensation()));
```
rosiereflo, line 138: long line!

sources/Inputs/include/Behaviours.h
```diff
@@ -72,5 +73,43 @@ struct Generator {
...
+      voltageRegulationOn{voltageRegulationOn},
+      hvdcLine{hvdcLine} {}
+  /**
+   * @brief Determines if two converter interfaces are equal
+   *
+   * @param other the other converter to be compared against
+   * 
+   * @return status of the comparaison
+   */
+  bool operator==(const ConverterInterface& other) const {
```
lecourtoisflo, line 105: Are you sure of this syntax for the equality operator ? I only know the version with the protoype:
bool operator==(const &, const &)

sources/Inputs/include/Behaviours.h
```diff
@@ -72,5 +73,43 @@ struct Generator {
...
+    return converterId == other.converterId && busId == other.busId && voltageRegulationOn == other.voltageRegulationOn && hvdcLine == other.hvdcLine;
+  }
+
+  ConverterInterfaceId converterId;                   ///< converter id
+  BusId busId;                                        ///< bus id
+  boost::optional<bool> voltageRegulationOn;          ///< voltage regulation parameter, for VSC converters only
+  boost::shared_ptr<dfl::inputs::HvdcLine> hvdcLine;  ///< hvdc line this converter is contained into
```
lecourtoisflo, line 112: style: no need for "dfl::inputs::"

sources/Inputs/include/Behaviours.h
```diff
@@ -72,5 +73,43 @@ struct Generator {
   double pmax;                             ///< maximum active power
 };
 
+class HvdcLine;
```
rosiereflo, line 76: Isn't it possible to avoid those 2 classes to reference each-other? It is always good to avoid such patterns

sources/Inputs/include/HvdcLine.h
```diff
@@ -0,0 +1,57 @@
...
+
+#pragma once
+
+#include "Behaviours.h"
+
+#include <boost/optional.hpp>
+#include <string>
+
+namespace dfl {
+/// @brief Namespace for inputs of Dynaflow launcher
```
lecourtoisflo, line 26: No need to duplicate this comment

sources/Inputs/include/NetworkManager.h
```diff
@@ -17,6 +17,8 @@
 
 #pragma once
 
+#include "HvdcLine.h"
+#include "NetworkManager.h"
```
lecourtoisflo, line 21: network manager MUST not include itself

sources/Inputs/include/Node.h
```diff
@@ -21,6 +21,7 @@
 
 #include <memory>
 #include <string>
+#include <unordered_map>
```
lecourtoisflo, line 24: not used

sources/Inputs/include/HvdcLine.h
```diff
@@ -0,0 +1,57 @@
...
+   * @param id the hvdc line id
+   * @param converterType type of converter of the hvdc line  
+   * @param converter1 first converter connected to the hvdc line
+   * @param converter2 second converter connected to the hvdc line
+   */
+  HvdcLine(const std::string& id, const ConverterType converterType, inputs::ConverterInterface converter1, inputs::ConverterInterface converter2);
+
+  const HvdcLineId id;                    ///< HvdcLine id
+  const ConverterType converterType;      ///< type of converter
+  inputs::ConverterInterface converter1;  ///< first converter
```
rosiereflo, line 53: Maybe we should avoid the copy here

sources/Inputs/include/HvdcLine.h
```diff
@@ -0,0 +1,57 @@
...
+
+  /**
+   * @brief Constructor
+   *
+   * @param id the hvdc line id
+   * @param converterType type of converter of the hvdc line  
+   * @param converter1 first converter connected to the hvdc line
+   * @param converter2 second converter connected to the hvdc line
+   */
+  HvdcLine(const std::string& id, const ConverterType converterType, inputs::ConverterInterface converter1, inputs::ConverterInterface converter2);
```
rosiereflo, line 49: const inputs::ConverterInterface&

sources/Inputs/src/NetworkManager.cpp
```diff
@@ -140,6 +144,34 @@ NetworkManager::buildTree() {
       LOG(debug) << "Node " << bus1->getID() << " connected to " << bus2->getID() << " and " << bus3->getID() << " by 3W " << (*it_t)->getID() << LOG_ENDL;
     }
   }
+
+  const auto& hvdcLines = network->getHvdcLines();
+  for (auto hvdcLine : hvdcLines) {
+    auto converter_dyn_1 = hvdcLine->getConverter1();
+    auto converter_1 = ConverterInterface(converter_dyn_1->getID(), converter_dyn_1->getBusInterface()->getID());
```
lecourtoisflo, line 151: Maybe check that the buses are connected. @marcochiaramello , could you confirm ?

sources/Outputs/include/Dyd.h
```diff
@@ -19,6 +19,7 @@
 
 #include "Algo.h"
 
+#include <DYDDynamicModelsCollection.h>
```
lecourtoisflo, line 22: Not formatted. Thie header should appear after DYDBlackBoxModel.h

sources/Inputs/include/Behaviours.h
rosiereflo: For example we could have the ids of the converters and their bus ids?

sources/Outputs/include/Dyd.h
```diff
@@ -74,20 +75,24 @@ class Dyd {
...
-        slackNode{slacknode} {}
+        slackNode{slacknode},
+        hvdcLines{hvdcLines} {}
 
     std::string basename;                               ///< basename for file
     std::string filename;                               ///< filepath for file to write
     std::vector<algo::GeneratorDefinition> generators;  ///< generators found
     std::vector<algo::LoadDefinition> loads;            ///< list of loads
     std::shared_ptr<inputs::Node> slackNode;            ///< slack node to use
+    std::vector<algo::HvdcLineDefinition> hvdcLines;    ///< list of hvdc lines
```
rosiereflo, line 95: I just noticed that but do we really need to copy all those vectors? cannot we use a reference?

sources/Outputs/include/Dyd.h
```diff
@@ -123,6 +128,16 @@ class Dyd {
    */
   static boost::shared_ptr<dynamicdata::BlackBoxModel> writeGenerator(const algo::GeneratorDefinition& def, const std::string& basename);
 
+  /**
+   * @brief Create black box model for hvdc line
+   *
+   * @param hvdcLine generator definition to use
+   * @param basename basename for file
+   *
+   * @returns black box model for generator
```
rosiereflo, line 137: generator->hvdc line

sources/Outputs/include/Dyd.h
```diff
@@ -175,8 +190,17 @@ class Dyd {
...
+   *   *
+   * @param collection the collection where the connection will be added
+   * @param hvdcLine the hvdc line definition to process
+   */
+  static void writeHvdcLineConnect(const boost::shared_ptr<dynamicdata::DynamicModelsCollection>& collection, const algo::HvdcLineDefinition& hvdcLine);
+
  private:
-  static const std::unordered_map<algo::GeneratorDefinition::ModelType, std::string>
+  static const std::unordered_map<algo::GeneratorDefinition::ModelType,
+                                  std::string>
```
rosiereflo, line 203: please put it back on 1 line

sources/Outputs/include/Par.h
```diff
@@ -45,20 +45,23 @@ class Par {
...
         filepath{filename},
         generators{gens},
+        hdvcLines{hdvcLines},
         activePowerCompensation{activePowerCompensation} {}
 
     std::string basename;                                                         ///< basename
     std::string dirname;                                                          ///< Dirname of output file relative to execution dir
     std::string filepath;                                                         ///< file path of the output file to write
     std::vector<algo::GeneratorDefinition> generators;                            ///< list of generators
+    std::vector<algo::HvdcLineDefinition> hdvcLines;                              ///< list of hdvc lines
```
rosiereflo, line 64: Same as above

sources/Outputs/include/Dyd.h
lecourtoisflo, line 203: This should not be formatted that way. I put the whole line in once and asked for formatting and the previous version was correct

sources/Outputs/src/Dyd.cpp
```diff
@@ -231,5 +267,15 @@ Dyd::writeGenConnect(const algo::GeneratorDefinition& def, unsigned int index) {
...
 
+void
+Dyd::writeHvdcLineConnect(const boost::shared_ptr<dynamicdata::DynamicModelsCollection>& collection, const algo::HvdcLineDefinition& hvdcLine) {
+  if (hvdcLine.position == algo::HvdcLineDefinition::Position::FIRST_IN_MAIN_COMPONENT) {
+    collection->addConnect("NETWORK", hvdcLine.converter1.busId + "_ACPIN", hvdcLine.id, "hvdc_terminal1");
+    collection->addConnect("NETWORK", hvdcLine.converter2.busId + "_ACPIN", hvdcLine.id, "hvdc_terminal2");
+  } else if (hvdcLine.position == algo::HvdcLineDefinition::Position::SECOND_IN_MAIN_COMPONENT) {
+    collection->addConnect("NETWORK", hvdcLine.converter1.busId + "_ACPIN", hvdcLine.id, "hvdc_terminal2");
+    collection->addConnect("NETWORK", hvdcLine.converter2.busId + "_ACPIN", hvdcLine.id, "hvdc_terminal1");
+  }
```
lecourtoisflo, line 278: Add a comment for the else case, for readability

lecourtoisflo (changes requested):

rosiereflo (changes requested):

sources/Inputs/include/Behaviours.h
PerninGuillaume: I verified in debug that the function was indeed called with a syntax like ConverterInterface == ConverterInterface2

sources/Outputs/include/Dyd.h
PerninGuillaume: I tried to reformat with CTRL+I or with CTRL+s and both put it in 3 differents lines, like before, did you have the comment when you tested it ?

sources/Outputs/src/Dyd.cpp
PerninGuillaume: What do I put in this else case ?

sources/Algo/src/Algo.cpp
PerninGuillaume: Yes

sources/Outputs/src/Dyd.cpp
lecourtoisflo: nothing just a comment or a TODO for future devs

sources/Outputs/include/Dyd.h
lecourtoisflo: Put all in one line and then reformat it

sources/Inputs/include/Behaviours.h
PerninGuillaume: Are you talking about the dfl::inputs::HvdcLine or the dfl::algo::HvdcLineDefinition ?

sources/Inputs/include/Behaviours.h
lecourtoisflo: OK then

sources/Algo/include/Algo.h
lecourtoisflo: also not formatted (comments are not aligned)

sources/Algo/src/Algo.cpp
rosiereflo: @cossartque Are we sure we won't have failures on some testcases that we didn't have earlier, or we didn't use testcases with HVDC at all ?

sources/Inputs/include/Behaviours.h
rosiereflo: HvdcLine and Converters

PerninGuillaume: For dfl::algo::HvdcLineDefinition, not factorizing the attributes would result in having in having 6 attributes compared to the 2 before, do we want that or a new structure ?

rosiereflo: it would be conerter ids, bus ids and?

sources/Inputs/src/NetworkManager.cpp
marcochiaramello: indeed, if converters are not connected, we do not need to specify anything in dyd

sources/Inputs/include/Behaviours.h
PerninGuillaume: And the voltageRegulationOn for the .par

sources/Inputs/src/NetworkManager.cpp
PerninGuillaume: So, if there are one of the converter of the hvdcLine which is in a bus that is not connected, I must skip this hvdc line in this for loop ?

sources/Inputs/src/NetworkManager.cpp
```diff
@@ -163,8 +163,7 @@ NetworkManager::buildTree() {
     } else {
       converterType = HvdcLine::ConverterType::LCC;
     }
-    auto hvdcLineCreated =
-        boost::make_shared<dfl::inputs::HvdcLine>(hvdcLine->getID(), converterType, converter_1, converter_2);
+    auto hvdcLineCreated = boost::make_shared<dfl::inputs::HvdcLine>(hvdcLine->getID(), converterType, converter_1, converter_2);
```
lecourtoisflo, line 166: why use boost shared_ptr here ? it is better to use std::shared_ptr when we can

sources/Outputs/include/Dyd.h
PerninGuillaume: I just tried to put all these vector as const vector& and the test passed, should I add it or drop it ? So we can do this for more attributes but in a separate PR that would fix a newly created issue ?

lecourtoisflo (changes requested):

sources/Outputs/include/Dyd.h
lecourtoisflo: It is working only because we call the function write of DydWriter in the same scope as the references variable. I think this is dangerous. If we want to avoid copy, the best way would be to use universal references and move

PerninGuillaume: Yes but where else would we call the method write of DydWriter ?

sources/Inputs/src/NetworkManager.cpp
PerninGuillaume: I used boost shared_ptr because I though it was the norm in this project, is it not ?

sources/Outputs/include/Dyd.h
rosiereflo: we cannot move as the same vectors are used in Par right?

sources/Inputs/include/Behaviours.h
rosiereflo: on my side I am ok with adding the attributes

PerninGuillaume: Or I could add a new Structure in dfl::algo : ConverterDefinition ?

sources/Outputs/include/Dyd.h
lecourtoisflo: correct, we cannot move. I don't like the idea of referencing local variables because we break the data encapsulation for a struct

sources/Inputs/src/NetworkManager.cpp
lecourtoisflo: No, it is a requirement from dynawo because we don't have access to c++11 in dynawo so we can use only shared pointers from boost. But in DFL, we are in c++11

lecourtoisflo: I already use shared_ptr in other places in DFL code (see Algo)

sources/Outputs/include/Dyd.h
PerninGuillaume: However, it could be a serious gain in time

sources/Inputs/include/Behaviours.h
rosiereflo: yes why not

lecourtoisflo: multiple encapsulation levels are not always a good idea. In that case, it is not a bug deal that it has more attributes. You can put in comments its origins or not if it iis obvious enough by the constructor

sources/Outputs/include/Dyd.h
lecourtoisflo: I would like a benchmark on this one. If there is a clear gain, OK, let's do it. If not, I would prefer we don't do this, since it breaks the original "spirit" of the structure

PerninGuillaume: If you benchmark it, change also the vector of the par

lecourtoisflo: Gain tested on Marco's IIDM is around 1ms (worst difference) to 3ms, with sometimes the version without the references better than the version with references. @rosiereflo whet do you think ?

PerninGuillaume: How much percentage gained is this 1ms ?

sources/Inputs/include/Behaviours.h
PerninGuillaume: Ok, this is done

sources/Algo/src/Algo.cpp
```diff
@@ -171,16 +171,26 @@ ControllerInterfaceDefinitionAlgorithm::operator()(const NodePtr& node) {
...
+    hvdcLines_.emplace_back(hvdcLine->id, hvdcLine->converterType, hvdcLine->converter1_id, hvdcLine->converter1_busId,
+                            hvdcLine->converter1_voltageRegulationOn, hvdcLine->converter2_id, hvdcLine->converter2_busId,
+                            hvdcLine->converter2_voltageRegulationOn, position);
   }
 }
+/////////////////////////////////////////////////////////////////
+
+bool
+HvdcLineDefinition::operator==(const HvdcLineDefinition& other) const {
+  return id == other.id && converterType == other.converterType && converter1_id == other.converter1_id && converter1_busId == other.converter1_busId &&
```
rosiereflo, line 191: Isn't this the default implementation of the equality?

rosiereflo (changes requested):

sources/Outputs/include/Dyd.h
rosiereflo: this is another topic so maybe we can discuss about this in another ticket.

sources/Algo/src/Algo.cpp
PerninGuillaume: [There are no default implementation of equality](https://stackoverflow.com/questions/217911/why-dont-c-compilers-define-operator-and-operator)

rosiereflo: oh right :)

PerninGuillaume: But it would be nice ;)

sources/Inputs/src/NetworkManager.cpp
cossartque: What do you mean by "not connected" @lecourtoisflo ?

sources/Algo/src/Algo.cpp
cossartque: For the moment, if both terminals are in the main cc, we do nothing, and dynawo uses the basic cpp model for hvdc and it works fine (in terms of robustness).
In the future, we will apply the proper Modelica model to improve the results quality but I wanted to do it in a different PR to avoid too large PR.

cossartque: So for the cases with HVDC that worked before, nothing will change (for the moment)
For the cases with HVDC that didn't work because they have one bus in the main cc and the other in a minor cc, they should work now.
For the cases with HVDC entirely in a minor cc, nothing will change and they normally should work as well

tests/main/res/config_hvdc_line_inverse.json
```diff
@@ -0,0 +1,7 @@
+{
+  "dfl-config": {
```
cossartque, line 2: This inverse test case should be private as it comes from a CGMES file. @marcochiaramello @rosiereflo  what do you think ?
I think we should remove it from the tests in the public repo and just do the inverse trick in the other case (we can inverse the lcc or the vsc). This way, in the same case we will have the normal configuration for one line and the inverse one for the other line.

cossartque (changes requested):

sources/Algo/src/Algo.cpp
rosiereflo: so if my understanding is correct we shouldn't fail if the hvdc is fully included in the main cc but just do nothing?

cossartque: Yes that's it

tests/main/res/config_hvdc_line_inverse.json
cossartque: @PerninGuillaume Can you add as a test the last iidm that I sent to you (the one with one vsc and one lcc) and inverse one of the two lines (for example you can write converterStation1="VSCStation2" converterStation2="VSCStation1" instead of converterStation1="VSCStation1" converterStation2="VSCStation2")

tests/main/CMakeLists.txt
```diff
@@ -54,6 +54,8 @@ DEFINE_LAUNCH_TEST(hvdc_line_normal)
 
 DEFINE_LAUNCH_TEST(hvdc_line_inverse)
```
cossartque, line 55: Thanks for adding the hvdc_dangling case.
I think we should remove the hvdc_line_inverse one as it should be private.

tests/main/res/TestIIDM_hvdc_dangling.iidm
```diff
@@ -0,0 +1,72 @@
...
+    <currentLimits1 permanentLimit="836.74000000000001"/>
+  </line>
+  <line id="_BUS____5-BUS___10-1_AC" name="BUS    5-BUS   10-1" r="0.01" x="0.050000000000000003" g1="0" b1="0" g2="0" b2="0" voltageLevelId1="_BUS____5_VL" bus1="_BUS____5_TN" connectableBus1="_BUS____5_TN" voltageLevelId2="_BUS___10_VL" bus2="_BUS___10_TN" connectableBus2="_BUS___10_TN" p1="400.03884322185792" q1="0.050012760310907206" p2="-400.02884066976839" q2="1.3706019826031923e-10">
+    <currentLimits1 permanentLimit="4183.6999999999998"/>
+  </line>
+  <hvdcLine id="HVDCLCCLine" r="0.02" nominalV="400" convertersMode="SIDE_1_INVERTER_SIDE_2_RECTIFIER" activePowerSetpoint="100" maxP="2000" converterStation1="LCCStation1" converterStation2="LCCStation2"/>
+  <hvdcLine id="HVDCVSCLine" r="0.02" nominalV="400" convertersMode="SIDE_1_INVERTER_SIDE_2_RECTIFIER" activePowerSetpoint="100" maxP="2000" converterStation1="VSCStation1" converterStation2="VSCStation2"/>
```
cossartque: Can you inverse one line ?
You can write converterStation1="VSCStation2" converterStation2="VSCStation1" instead of converterStation1="VSCStation1" converterStation2="VSCStation2"

cossartque (changes requested):

sources/Inputs/src/NetworkManager.cpp
lecourtoisflo: elemenst (generators, loads, etc...) have an API getInitialConnected which determines if they are connected to the network. Since we are using it for generators, I thought it would be a good idea to use it for converters

tests/main/CMakeLists.txt
PerninGuillaume: Should I wait for confirmation before removing the test for @rosiereflo or @marcochiaramello  ?

cossartque: If we have a complete test with both vsc and lcc, both normal and inverse (you just have to inverse the converterStations of one of the two lines), then there is no need for this one.
But let's wait for @marcochiaramello and @rosiereflo confirmation.

tests/inputs/res/HvdcDangling.iidm
```diff
@@ -0,0 +1,72 @@
...
+    <currentLimits1 permanentLimit="836.74000000000001"/>
+  </line>
+  <line id="_BUS____5-BUS___10-1_AC" name="BUS    5-BUS   10-1" r="0.01" x="0.050000000000000003" g1="0" b1="0" g2="0" b2="0" voltageLevelId1="_BUS____5_VL" bus1="_BUS____5_TN" connectableBus1="_BUS____5_TN" voltageLevelId2="_BUS___10_VL" bus2="_BUS___10_TN" connectableBus2="_BUS___10_TN" p1="400.03884322185792" q1="0.050012760310907206" p2="-400.02884066976839" q2="1.3706019826031923e-10">
+    <currentLimits1 permanentLimit="4183.6999999999998"/>
+  </line>
+  <hvdcLine id="HVDCLCCLine" r="0.02" nominalV="400" convertersMode="SIDE_1_INVERTER_SIDE_2_RECTIFIER" activePowerSetpoint="100" maxP="2000" converterStation1="LCCStation1" converterStation2="LCCStation2"/>
+  <hvdcLine id="HVDCVSCLine" r="0.02" nominalV="400" convertersMode="SIDE_1_INVERTER_SIDE_2_RECTIFIER" activePowerSetpoint="100" maxP="2000" converterStation1="VSCStation1" converterStation2="VSCStation2"/>
```
cossartque, line 71: Can you also invert here ?

tests/main/CMakeLists.txt
marcochiaramello: yes, I think we can remove that case

cossartque: In fact I think you don't even need the normal test case either but just the hvdc_dangling that now covers everything

PerninGuillaume: It does not hurt to have a test that could cover cases we did not know about

rosiereflo: We should have one or 2 tests with the 4 possibilities inside (VSC-LCC)/(SIDE1-SIDE2). As long as this is covered I will be OK

tests/inputs/res/HvdcDangling.iidm
PerninGuillaume: Done

tests/main/CMakeLists.txt
cossartque: The hvdc_dangling covers them all as Guillaume did the inversion of one of the two lines

rosiereflo:
@reviewers: Could you mark as resolved the conversations that were addressed? Otherwise it becomes hard to follow

sources/Inputs/include/Behaviours.h
```diff
@@ -72,5 +73,43 @@ struct Generator {
...
+
+  /**
+   * @brief Constructor
+   *
+   * @param converterId the id of the converter
+   * @param busId the id of the bus
+   * @param voltageRegulationOn optional boolean for the voltage regulation parameter, only used for VSC converters
+   * @param hvdcLine the hvdc line this converter is contained into
+   */
+  explicit ConverterInterface(const ConverterInterfaceId& converterId, BusId busId, boost::optional<bool> voltageRegulationOn = {},
```
lecourtoisflo, line 92: why put default values for arguments when it is not used ?

sources/Inputs/include/HvdcLine.h
```diff
@@ -0,0 +1,77 @@
...
+   * @param converterType type of converter of the hvdc line  
+   * @param converter1_id first converter id 
+   * @param converter1_busId first converter bus id
+   * @param converter1_voltageRegulationOn firt converter voltage regulation parameter, for VSC converters only
+   * @param converter2_id second converter id 
+   * @param converter2_busId second converter bus id
+   * @param converter2_voltageRegulationOn second converter voltage regulation parameter, for VSC converters only
+   */
+  HvdcLine(const std::string& id, const ConverterType converterType, const ConverterId& converter1_id, const BusId& converter1_busId,
+           const boost::optional<bool> converter1_voltageRegulationOn, const ConverterId& converter2_id, const BusId& converter2_busId,
```
lecourtoisflo, line 64: optional are structs so it is better to use const& for copy

sources/Inputs/src/HvdcLine.cpp
```diff
@@ -0,0 +1,40 @@
...
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at http://mozilla.org/MPL/2.0/.
+// SPDX-License-Identifier: MPL-2.0
+//
+
+/**
+ * @file  Node.cpp
+ *
+ * @brief Node structure implementation file
```
lecourtoisflo, line 14: Incorrect comments

sources/Inputs/src/NetworkManager.cpp
```diff
@@ -140,6 +144,38 @@ NetworkManager::buildTree() {
...
+
+      auto vsc_converter_dyn_2 = boost::static_pointer_cast<DYN::VscConverterInterface>(converter_dyn_2);
+      converter_2.voltageRegulationOn = vsc_converter_dyn_2->getVoltageRegulatorOn();
+    } else {
+      converterType = HvdcLine::ConverterType::LCC;
+    }
+    auto hvdcLineCreated =
+        std::make_shared<dfl::inputs::HvdcLine>(hvdcLine->getID(), converterType, converter_1.converterId, converter_1.busId, converter_1.voltageRegulationOn,
+                                                converter_2.converterId, converter_2.busId, converter_2.voltageRegulationOn);
+    hvdcLines_.emplace_back(hvdcLineCreated);
```
lecourtoisflo, line 172: Add a debug log like other components (for consistency)

lecourtoisflo (changes requested):

sources/Inputs/src/NetworkManager.cpp
```diff
@@ -21,14 +21,18 @@
...
+#include <DYNHvdcLineInterface.h>
+#include <DYNLccConverterInterface.h>
 #include <DYNLineInterface.h>
 #include <DYNLoadInterface.h>
 #include <DYNNetworkInterface.h>
 #include <DYNThreeWTransformerInterface.h>
 #include <DYNTwoWTransformerInterface.h>
 #include <DYNVoltageLevelInterface.h>
+#include <DYNVscConverterInterface.h>
 #include <boost/make_shared.hpp>
```
lecourtoisflo, line 36: Remove all references to boost make_shared now that you are using std::shared_ptr

lecourtoisflo (changes requested):

sources/Algo/src/Algo.cpp
rosiereflo: then we need an update here

sources/Inputs/include/Behaviours.h
PerninGuillaume: It is used in the networkmanager.cpp, when we create the ConverterInterface we don't know yet the value, so for readability I put the voltageRegulationOn with a default value.
So we have 
`    auto converter_1 = ConverterInterface(converter_dyn_1->getID(), converter_dyn_1->getBusInterface()->getID());
`
instead of 
`    auto converter_1 = ConverterInterface(converter_dyn_1->getID(), converter_dyn_1->getBusInterface()->getID(), boost::optional<bool>());
`

sources/Algo/src/Algo.cpp
```diff
@@ -162,6 +162,35 @@ LoadDefinitionAlgorithm::operator()(const NodePtr& node) {
     loads_.emplace_back(it->id, node->id);
   }
 }
+/////////////////////////////////////////////////////////////////
+
+ControllerInterfaceDefinitionAlgorithm::ControllerInterfaceDefinitionAlgorithm(std::vector<HvdcLineDefinition>& hvdcLines) : hvdcLines_(hvdcLines) {}
+
+void
+ControllerInterfaceDefinitionAlgorithm::operator()(const NodePtr& node) {
+  for (auto converter : node->converterInterfaces) {
```
lecourtoisflo, line 171: better use auto& or const auto& here

sources/Algo/src/Algo.cpp
```diff
@@ -162,6 +162,35 @@ LoadDefinitionAlgorithm::operator()(const NodePtr& node) {
...
+    auto hvdcLine = converter.hvdcLine;
+    HvdcLineDefinition::Position position;
+    if (converter.converterId == hvdcLine->converter1_id) {
+      position = HvdcLineDefinition::Position::FIRST_IN_MAIN_COMPONENT;
+    } else if (converter.converterId == hvdcLine->converter2_id) {
+      position = HvdcLineDefinition::Position::SECOND_IN_MAIN_COMPONENT;
+    } else {
+      LOG(error) << MESS(HvdcLineBadInitialization, hvdcLine->id) << LOG_ENDL;
```
lecourtoisflo, line 179: I would prefer that you don't emplace the line if there is an error

tests/outputs/TestPar.cpp
```diff
@@ -12,6 +12,7 @@
 #include "Tests.h"
 
 #include <boost/filesystem.hpp>
+#include <boost/make_shared.hpp>
```
lecourtoisflo, line 15: do you need this header now ? same question for all unit tests

lecourtoisflo (changes requested):

lecourtoisflo (approved):

sources/Algo/src/Algo.cpp
```diff
@@ -168,7 +168,7 @@ ControllerInterfaceDefinitionAlgorithm::ControllerInterfaceDefinitionAlgorithm(s
 
 void
 ControllerInterfaceDefinitionAlgorithm::operator()(const NodePtr& node) {
-  for (auto converter : node->converterInterfaces) {
+  for (const auto& converter : node->converterInterfaces) {
     auto hvdcLine = converter.hvdcLine;
```
rosiereflo, line 172: const auto&

rosiereflo (changes requested):

sources/Algo/src/Algo.cpp
PerninGuillaume: Is it not ?

rosiereflo: on auto hvdcLine

tests/main/CMakeLists.txt
cossartque: @PerninGuillaume can you confirm that you removed the whole inverse case ? (it is a private case)

sources/Algo/src/Algo.cpp
cossartque: @PerninGuillaume has it been updated ?

tests/main/CMakeLists.txt
PerninGuillaume: Now, it is

sources/Algo/src/Algo.cpp
PerninGuillaume: Not yet, if you want to check the case of both converters in the main connex component, then I need to change a few things to be able to know when we are in this case.

PerninGuillaume: Done

sources/Algo/src/Algo.cpp
```diff
@@ -169,12 +169,22 @@ ControllerInterfaceDefinitionAlgorithm::ControllerInterfaceDefinitionAlgorithm(s
...
-      position = HvdcLineDefinition::Position::FIRST_IN_MAIN_COMPONENT;
+      if (!inserted) {
+        position = HvdcLineDefinition::Position::BOTH_IN_MAIN_COMPONENT;
+      } else {
+        position = HvdcLineDefinition::Position::FIRST_IN_MAIN_COMPONENT;
+      }
     } else if (converter.converterId == hvdcLine->converter2_id) {
-      position = HvdcLineDefinition::Position::SECOND_IN_MAIN_COMPONENT;
+      if (!inserted) {
+        position = HvdcLineDefinition::Position::BOTH_IN_MAIN_COMPONENT;
```
lecourtoisflo, line 184: With this code, we will add twice the same line with 2 different positions in hvdcLines_, correct ? Is this something we want ?

lecourtoisflo (changes requested):

sources/Algo/src/Algo.cpp
PerninGuillaume: No we don't want that, I will modify it

sources/Algo/include/Algo.h
```diff
@@ -259,5 +260,98 @@ class LoadDefinitionAlgorithm : public NodeAlgorithm {
...
+    SECOND_IN_MAIN_COMPONENT,  ///< the second converter of this hvdc line is in the main connex component
+    BOTH_IN_MAIN_COMPONENT     ///< both converters of this hvdc line are in the main connex component
+  };
+
+  /**
+   * @brief Determines if two converter interfaces are equal
+   *
+   * @param other the other converter to be compared against
+   * 
+   * @return status of the comparaison
```
marcochiaramello, line 288: comparison

sources/Algo/include/Algo.h
```diff
@@ -259,5 +260,98 @@ class LoadDefinitionAlgorithm : public NodeAlgorithm {
...
+   * and their position compared to the main connex component.
+   */
+  enum class Position {
+    FIRST_IN_MAIN_COMPONENT,   ///< the first converter of this hvdc line is in the main connex component
+    SECOND_IN_MAIN_COMPONENT,  ///< the second converter of this hvdc line is in the main connex component
+    BOTH_IN_MAIN_COMPONENT     ///< both converters of this hvdc line are in the main connex component
+  };
+
+  /**
+   * @brief Determines if two converter interfaces are equal
```
marcochiaramello, line 284: converters'

sources/Inputs/include/HvdcLine.h
```diff
@@ -0,0 +1,77 @@
...
+    VSC,  ///< voltage source converter
+    LCC   ///< line-commutated converter
+  };
+
+  /**
+   * @brief Determines if two converter interfaces are equal
+   *
+   * @param other the other converter to be compared against
+   * 
+   * @return status of the comparaison
```
marcochiaramello, line 47: comparison

sources/Inputs/include/HvdcLine.h
```diff
@@ -0,0 +1,77 @@
...
+  using BusId = std::string;        ///< alias for bus id
+
+  /// @brief Type of converter
+  enum class ConverterType {
+    VSC,  ///< voltage source converter
+    LCC   ///< line-commutated converter
+  };
+
+  /**
+   * @brief Determines if two converter interfaces are equal
```
marcochiaramello, line 43: converters'

marcochiaramello (changes requested):

sources/Inputs/include/HvdcLine.h
lecourtoisflo: For me we can keep "converter interfaces" as it is closer to the type name

marcochiaramello: ok

sources/Algo/include/Algo.h
```diff
@@ -323,20 +323,49 @@ struct HvdcLineDefinition {
...
+    return lhs.id < rhs.id;
+  }
+};
 /**
  * @brief the controller interface definition algorithm
  */
 class ControllerInterfaceDefinitionAlgorithm : public NodeAlgorithm {
  public:
+  using HvdcLineId = std::string;                              ///< HvdcLine id definition
+  using HvdcLineSet = std::set<HvdcLineDefinition, HashTerm>;  ///< TODO
```
rosiereflo, line 361: Do we need it to be sorted? Moreover a unordered map id->Definition might be better to avoid having to explicitly define operators?

PerninGuillaume: The issue I encountered when trying to make it unsorted is in the tests. With an unsorted container the order of the elements in a for loop is not always the same(I think). So if we write in the dyd file an hvdc before a lcc converter we will have an issue in the tests when we verify that the files are equal.

rosiereflo: ok, but we can still use a map to avoid heavy structures

PerninGuillaume: I don't get what will we gain from using a map ? Clarity of code ?

sources/Algo/src/Algo.cpp
```diff
@@ -164,35 +164,32 @@ LoadDefinitionAlgorithm::operator()(const NodePtr& node) {
...
   for (const auto& converter : node->converterInterfaces) {
     const auto& hvdcLine = converter.hvdcLine;
-    bool inserted;
-    std::tie(std::ignore, inserted) = hvdcLinesIds_.emplace(hvdcLine->id);
-    HvdcLineDefinition::Position position;
+    HvdcLineDefinition createdHvdcLine = HvdcLineDefinition(hvdcLine->id, hvdcLine->converterType, hvdcLine->converter1_id, hvdcLine->converter1_busId,
+                                                            hvdcLine->converter1_voltageRegulationOn, hvdcLine->converter2_id, hvdcLine->converter2_busId,
+                                                            hvdcLine->converter2_voltageRegulationOn, HvdcLineDefinition::Position::BOTH_IN_MAIN_COMPONENT);
+    auto it = hvdcLines_.find(createdHvdcLine);
+    bool alreadyInserted = it != hvdcLines_.end();
```
lecourtoisflo, line 177: you can check the iterator without using an local variable

sources/Algo/include/Algo.h
rosiereflo: yes

sources/Algo/include/Algo.h
```diff
@@ -323,20 +323,49 @@ struct HvdcLineDefinition {
...
+    return std::hash<std::string>{}(hvdcLine.id);
+  }
+  /**
+   * @brief Less function
+   *
+   * @param lhs the left hand side to compare
+   * @param rhs the right hand side to compare
+   * @returns true if lhs is inferior to rhs based on their id
+   */
+  bool operator()(const HvdcLineDefinition& lhs, const HvdcLineDefinition& rhs) const {
```
lecourtoisflo, line 351: this function should not be in the hash struct. It is the comparator function. Remove if not different from the already defined '<' operator

sources/Algo/include/Algo.h
```diff
@@ -323,20 +323,49 @@ struct HvdcLineDefinition {
   const ConverterId converter2_id;                             ///< second converter id
   const BusId converter2_busId;                                ///< second converter bus id
   const boost::optional<bool> converter2_voltageRegulationOn;  ///< second converter voltage regulation parameter, for VSC converters only
-  const Position position;                                     ///< position of the converters of this hvdc line compared to the main connex component
+  mutable Position position;                                   ///< position of the converters of this hvdc line compared to the main connex component
```
lecourtoisflo, line 326: mutable is only for const objects that must be modified. This is not the case here.

lecourtoisflo (changes requested):

sources/Algo/src/Algo.cpp
PerninGuillaume: It is for the clarity of the code

sources/Algo/include/Algo.h
PerninGuillaume: It was useful for the set of the find function, this function returned a const iterator and I needed to modify the position of the element pointed by the iterator. Hence the mutable.(thanks stackoverflow).
But with the map it won't be needed anymore as it does not return a const iterator

sources/Algo/include/Algo.h
lecourtoisflo: Also the problem of set is that we cannot modify the key inplace

sources/Algo/include/Algo.h
```diff
@@ -259,5 +260,101 @@ class LoadDefinitionAlgorithm : public NodeAlgorithm {
...
+};
+
+/**
+ * @brief the controller interface definition algorithm
+ */
+class ControllerInterfaceDefinitionAlgorithm : public NodeAlgorithm {
+ public:
+  using HvdcLineId = std::string;  ///< HvdcLine id definition
+  using HvdcLineMap =
+      std::map<HvdcLineId, HvdcLineDefinition>;  ///< HvdcLineMap definition, useful in algo.cpp for checking if a hvdcLineDefinition is already inserted
```
lecourtoisflo, line 336: innacurate comment: the map is the output of the algorithm and is provided by the context

sources/Inputs/src/HvdcLine.cpp
```diff
@@ -0,0 +1,40 @@
...
+    converterType{converterType},
+    converter1_id{converter1_id},
+    converter1_busId{converter1_busId},
+    converter1_voltageRegulationOn{converter1_voltageRegulationOn},
+    converter2_id{converter2_id},
+    converter2_busId{converter2_busId},
+    converter2_voltageRegulationOn{converter2_voltageRegulationOn} {}
+
+bool
+HvdcLine::operator==(const HvdcLine& other) const {
```
lecourtoisflo, line 34: To be consistent with the further operators for algorithm definitions, define the equality operator only using the id

sources/Inputs/src/NetworkManager.cpp
```diff
@@ -140,6 +144,40 @@ NetworkManager::buildTree() {
...
+    if (!converter_dyn_1->getInitialConnected() || !converter_dyn_2->getInitialConnected()) {
+      continue;
+    }
+    auto converter_1 = ConverterInterface(converter_dyn_1->getID(), converter_dyn_1->getBusInterface()->getID());
+    auto converter_2 = ConverterInterface(converter_dyn_2->getID(), converter_dyn_2->getBusInterface()->getID());
+
+    HvdcLine::ConverterType converterType;
+    if (converter_dyn_1->getConverterType() == DYN::ConverterInterface::ConverterType_t::VSC_CONVERTER) {
+      converterType = HvdcLine::ConverterType::VSC;
+      auto vsc_converter_dyn_1 = boost::static_pointer_cast<DYN::VscConverterInterface>(converter_dyn_1);
```
lecourtoisflo, line 161: Since we are using pointers, it is better to use dynamic_pointer_cast here to check the inheritance

lecourtoisflo (changes requested):

sources/Inputs/src/NetworkManager.cpp
PerninGuillaume: I am not sure why. If we are sure of the underlying type is it not better to use the static_pointer_cast ?

sources/Algo/include/Algo.h
PerninGuillaume: So what should I write ?

sources/Inputs/src/HvdcLine.cpp
PerninGuillaume: Even if it is not used ?

sources/Inputs/src/NetworkManager.cpp
lecourtoisflo: static cast will always work since we are dealing with pointers, while dynamic cast will not

sources/Inputs/src/HvdcLine.cpp
lecourtoisflo: if it is not used, remove it

sources/Algo/include/Algo.h
lecourtoisflo: generic comment, like 'Alias for map of hvdc line definition"

sources/Inputs/src/NetworkManager.cpp
rosiereflo: dynamimc cast is safer as it makes a type check but less efficient performance wise. However we call this once so maybe using dynamic cast is better indeed

lecourtoisflo: once by hdvc line

rosiereflo: yes, let me rephrase: I meant that this algorithm is called once at the beginning of the flow and Hvdc lines are not numerous so it should be OK

PerninGuillaume: But we check it beforehand with the if condition so we are sure this cast will suceed, that's exactly when we want to use a static pointer cast

lecourtoisflo: OK

sources/Inputs/src/HvdcLine.cpp
PerninGuillaume: It is used in the tests, and it needs to test every member to be exhaustive. Maybe you want me to rename the function ?

lecourtoisflo: If you need in the tests, implement this operator or equivalent only in tests

sources/Inputs/src/NetworkManager.cpp
lecourtoisflo: The check before is independant from the type of the object

PerninGuillaume: How is it independent ? The if condition checks if the type of the object is VSC_CONVERTER

lecourtoisflo: The type attribute and the type of the object are independant. You can imagine a bug where the wrong type attribute is given to a structure type

PerninGuillaume: If we don't trust the code that has already been tested a bit, then I can not imagine the number of tests we would put every 10 lines of code.

lecourtoisflo: OK

lecourtoisflo (approved):

cossartque (approved):

rosiereflo (approved):

sources/Inputs/include/Behaviours.h
```diff
@@ -72,5 +73,43 @@ struct Generator {
...
+      converterId{converterId},
+      busId{busId},
+      voltageRegulationOn{voltageRegulationOn},
+      hvdcLine{hvdcLine} {}
+  /**
+   * @brief Determines if two converter interfaces are equal
+   *
+   * @param other the other converter to be compared against
+   * 
+   * @return status of the comparaison
```
marcochiaramello, line 103: comparison

marcochiaramello (changes requested):

rosiereflo:
Is the draft status still valid?

PerninGuillaume:
Uh, I tried removing the line in the cmakelist in dynawo, recompiled and it still seemed to work. So maybe the draft status is no more valid, but to be sure, one could try to checkout this branch and compile to see if it really works.

rosiereflo:
I meant the PR draft status

PerninGuillaume:
We could merge this, if I were to be sure that the line in the cmakelist needs not to be added

PerninGuillaume:
No, finally florent checked out this branch, and it did not compile without the line in the cmakeList

marcochiaramello (approved):

cossartque:
@rosiereflo @marcochiaramello I propose to merge the dev, is that ok for you ?

PR #93, opened 2020-12-15
Use dynawo service manager to manage generators connected by switches

### LINKED ISSUE #79: [iidm 1.2] consider retained switch when determing main topological island

### LINKED ISSUE #80: [iidm 1.2] use of generator with line impedence model
When multiple generators (in voltage regulation) are attached to two nodes connected via a retained switch, the generator with line impedence model should be used to avoid twice the same equation in the system

lecourtoisflo:
closes #80
closes #79

depends on dynawo PR 1346

### DIFF
--- a/sources/Algo/src/Algo.cpp
+++ b/sources/Algo/src/Algo.cpp
@@ -74,19 +74,28 @@ MainConnexComponentAlgorithm::operator()(const NodePtr& node) {
 
 ////////////////////////////////////////////////////////////////
 
-GeneratorDefinitionAlgorithm::GeneratorDefinitionAlgorithm(Generators& gens, bool infinitereactivelimits) :
+GeneratorDefinitionAlgorithm::GeneratorDefinitionAlgorithm(Generators& gens, bool infinitereactivelimits,
+                                                           const boost::shared_ptr<DYN::ServiceManagerInterface>& serviceManager) :
     NodeAlgorithm(),
     generators_(gens),
-    useInfiniteReactivelimits_{infinitereactivelimits} {}
+    useInfiniteReactivelimits_{infinitereactivelimits},
+    serviceManager_(serviceManager) {}
 
 void
 GeneratorDefinitionAlgorithm::operator()(const NodePtr& node) {
   auto& node_generators = node->generators;
   if (node_generators.size() == 1) {
     auto model = useInfiniteReactivelimits_ ? GeneratorDefinition::ModelType::SIGNALN : GeneratorDefinition::ModelType::DIAGRAM_PQ_SIGNALN;
+    if (IsOtherGeneratorConnectedBySwitches(node)) {
+      // Use the model with impedance in case another generator is connected to the node through a switch network path
+      model = useInfiniteReactivelimits_ ? GeneratorDefinition::ModelType::WITH_IMPEDANCE_SIGNALN
+                                         : GeneratorDefinition::ModelType::WITH_IMPEDANCE_DIAGRAM_PQ_SIGNALN;
+    }
+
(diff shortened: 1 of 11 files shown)


sources/Inputs/src/Node.cpp
```diff
@@ -19,7 +19,19 @@
 
 namespace dfl {
 namespace inputs {
-Node::Node(const NodeId& idNode, double nominalVoltageNode) : id(idNode), nominalVoltage{nominalVoltageNode}, neighbours{} {}
+
+std::shared_ptr<Node>
+Node::build(const NodeId& id, const std::shared_ptr<VoltageLevel>& vl, double nominalVoltage) {
+  auto ret = std::shared_ptr<Node>(new Node(id, vl, nominalVoltage));
```
PerninGuillaume, line 25: Why not use std::make_shared ?

sources/Algo/src/Algo.cpp
```diff
@@ -148,6 +157,29 @@ GeneratorDefinitionAlgorithm::isDiagramValid(const inputs::Generator& generator)
   return valid;
 }
 
+bool
+GeneratorDefinitionAlgorithm::IsOtherGeneratorConnectedBySwitches(const NodePtr& node) const {
+  auto vl = node->voltageLevel.lock();
```
PerninGuillaume, line 162: Are we sure this lock will give us back a valid shared pointer ?

lecourtoisflo: At this point, yes because life cycle of the nodes is the same as the voltage levels (Network manager)

sources/Inputs/src/Node.cpp
lecourtoisflo: std::make_shared needs the constructor to be public (maybe in higher standard it is not the case)

tests/algo/TestAlgo.cpp
```diff
@@ -21,13 +21,70 @@
...
+  void clear() {
+    map_.clear();
+  }
+
+  void add(const std::string& busId, const std::string& VLId, const std::string& value) {
+    map_[std::tie(busId, VLId)].push_back(value);
+    map_[std::tie(value, VLId)].push_back(busId);
+  }
+
+  std::vector<std::string> getBusesConnectedBySwitch(const std::string& busId, const std::string& VLId) const final {
```
rosiereflo, line 45: Missing comments

rosiereflo (changes requested):

sources/Inputs/src/NetworkManager.cpp
```diff
@@ -60,13 +61,16 @@ NetworkManager::buildTree() {
       std::exit(EXIT_FAILURE);
     }
 
+    auto vl = std::make_shared<VoltageLevel>((*it_v)->getID());
+    voltagelevels_.push_back(vl);
```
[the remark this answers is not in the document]
lecourtoisflo: OK

tests/algo/TestAlgo.cpp
lecourtoisflo: OK

sources/Inputs/src/NetworkManager.cpp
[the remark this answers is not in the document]
rosiereflo: *vector but I guess you got my point :)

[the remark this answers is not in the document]
lecourtoisflo: This vector is used only for the storage of shared_ptr. They are always used via the weak_ptr in nodes

rosiereflo (approved):

marcochiaramello (approved):

Issue #87, opened 2020-12-03
Remove generators with targetP = 0 from the active power compensation


cossartque:
Generators that have targetP = 0 but that still participate in the voltage regulation must be removed from the active power compensation.
As a result they should have KGover = 0;
@adriguir fyi

cossartque:
@PerninGuillaume this issue can be closed as well ?

PR #94, opened 2020-12-17
Target_ajustements: initial commit

### LINKED ISSUE #84: Use the simple cpp generator model when targetP < pMin or targetP > pMax
We have seen cases in examples where in the iidm targetP (or p) is lower than pMin or higher than pMax.
In that case we should not implement a generator model in the .dyd file but do nothing and let dynawo use a simple injector cpp model.
@marcochiaramello what do you think ?

PerninGuillaume:
closes #84 #87 

### DIFF
--- a/sources/Inputs/src/NetworkManager.cpp
+++ b/sources/Inputs/src/NetworkManager.cpp
@@ -103,7 +103,7 @@ NetworkManager::buildTree() {
       auto pmin = (*it_g)->getPMin();
       auto pmax = (*it_g)->getPMax();
       if ((*it_g)->isVoltageRegulationOn() && (DYN::doubleEquals(targetP, pmin) || targetP > pmin) && (DYN::doubleEquals(targetP, pmax) || targetP < pmax)) {
-        // We don't use dynamic models for generators with voltage regulation disabled
+        // We don't use dynamic models for generators with voltage regulation disabled and an active power reference outside the generator's PQ diagram
         nodes_[nodeid]->generators.emplace_back((*it_g)->getID(), (*it_g)->getReactiveCurvesPoints(), (*it_g)->getQMin(), (*it_g)->getQMax(),
                                                 pmin, pmax, targetP);
         LOG(debug) << "Node " << nodeid << " contains generator " << (*it_g)->getID() << LOG_ENDL;
--- a/sources/Inputs/src/NetworkManager.cpp
+++ b/sources/Inputs/src/NetworkManager.cpp
@@ -98,10 +98,11 @@ NetworkManager::buildTree() {
       // node should exist at this point
       assert(nodes_.count(nodeid));
 #endif
-      if ((*it_g)->isVoltageRegulationOn()) {
+      auto targetP = (*it_g)->getTargetP();
+      if ((*it_g)->isVoltageRegulationOn() && targetP >= (*it_g)->getPMin() && targetP <= (*it_g)->getPMax()) {
         // We don't use dynamic models for generators with voltage regulation disabled
         nodes_[nodeid]->generators.emplace_back((*it_g)->getID(), (*it_g)->getReactiveCurvesPoints(), (*it_g)->getQMin(), (*it_g)->getQMax(),
-                                                (*it_g)->getPMin(), (*it_g)->getPMax());
+                                                (*it_g)->getPMin(), (*it_g)->getPMax(), targetP);
         LOG(debug) << "Node " << nodeid << " contains generator " << (*it_g)->getID() << LOG_ENDL;
       }
     }
--- a/sources/Inputs/src/NetworkManager.cpp
+++ b/sources/Inputs/src/NetworkManager.cpp
@@ -21,6 +21,7 @@
 #include "Message.hpp"
 
 #include <DYNBusInterface.h>
+#include <DYNCommon.h>
 #include <DYNConverterInterface.h>
 #include <DYNDataInterfaceFactory.h>
 #include <DYNGeneratorInterface.h>
@@ -99,10 +100,12 @@ NetworkManager::buildTree() {
       assert(nodes_.count(nodeid));
 #endif
       auto targetP = (*it_g)->getTargetP();
-      if ((*it_g)->isVoltageRegulationOn() && targetP >= (*it_g)->getPMin() && targetP <= (*it_g)->getPMax()) {
+      auto pmin = (*it_g)->getPMin();
+      auto pmax = (*it_g)->getPMax();
+      if ((*it_g)->isVoltageRegulationOn() && (DYN::doubleEquals(targetP, pmin) || targetP > pmin) && (DYN::doubleEquals(targetP, pmax) || targetP < pmax)) {
         // We don't use dynamic models for generators with voltage regulation disabled
         nodes_[nodeid]->generators.emplace_back((*it_g)->getID(), (*it_g)->getReactiveCurvesPoints(), (*it_g)->getQMin(), (*it_g)->getQMax(),
-                                                (*it_g)->getPMin(), (*it_g)->getPMax(), targetP);
+                                                pmin, pmax, targetP);
         LOG(debug) << "Node " << nodeid << " contains generator " << (*it_g)->getID() << LOG_ENDL;
       }
     }
--- a/sources/Outputs/include/Par.h
+++ b/sources/Outputs/include/Par.h
@@ -95,7 +95,7 @@ class Par {
    *
    * @param set the parameter set to update
    * @param activePowerCompensation the type of active power compensation
-   * @param fixedP boolean to determines if the set represents a generator with a targetP equal to 0
+   * @param fixedP boolean to determine if the set represents a generator with a targetP equal to 0
    */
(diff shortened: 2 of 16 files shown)


sources/Inputs/src/NetworkManager.cpp
```diff
@@ -98,10 +98,11 @@ NetworkManager::buildTree() {
       // node should exist at this point
       assert(nodes_.count(nodeid));
 #endif
-      if ((*it_g)->isVoltageRegulationOn()) {
+      auto targetP = (*it_g)->getTargetP();
+      if ((*it_g)->isVoltageRegulationOn() && targetP >= (*it_g)->getPMin() && targetP <= (*it_g)->getPMax()) {
         // We don't use dynamic models for generators with voltage regulation disabled
```
cossartque, line 103: We don't use dynamic models for generators with voltage regulation disabled and an active power reference outside the generator's PQ diagram

sources/Inputs/src/NetworkManager.cpp
```diff
@@ -98,10 +98,11 @@ NetworkManager::buildTree() {
       // node should exist at this point
       assert(nodes_.count(nodeid));
 #endif
-      if ((*it_g)->isVoltageRegulationOn()) {
+      auto targetP = (*it_g)->getTargetP();
+      if ((*it_g)->isVoltageRegulationOn() && targetP >= (*it_g)->getPMin() && targetP <= (*it_g)->getPMax()) {
```
lecourtoisflo, line 102: Use dynawo API isDoubleEqual for equality comparaison

lecourtoisflo (changes requested):

tests/outputs/TestPar.cpp
```diff
@@ -35,31 +35,34 @@ TEST(TestPar, write) {
...
                                                                      {
                                                                          GeneratorDefinition::ReactiveCurvePoint(1., 11., 110.),
                                                                          GeneratorDefinition::ReactiveCurvePoint(2., 22., 220.),
                                                                          GeneratorDefinition::ReactiveCurvePoint(3., 33., 330.),
                                                                          GeneratorDefinition::ReactiveCurvePoint(4., 44., 440.),
                                                                      },
-                                                                     4., 40., 44., 440.)};
+                                                                     4., 40., 44., 440., 100.),
+                                                 GeneratorDefinition("G4", GeneratorDefinition::ModelType::DIAGRAM_PQ_SIGNALN, "04",
+                                                                     {},
```
cossartque, line 64: Can you also test the case of a generator with targetP = 0 and without a diagram ?
This way we could see that for generators without diagrams we now have two sets of parameters : one with KGover=1 and one with KGover=0 and that in the .dyd we have the proper par set for each generator

cossartque (changes requested):

lecourtoisflo (approved):

tests/outputs/TestPar.cpp
PerninGuillaume: Hum, I did not think of that, that means that for every constant set in the par (such as impSignalNGenerator or signalNGenerator) we will duplicate it with the only thing changing being the "generator_KGover" ?

cossartque: That's it yes !
We would have impSignalNGenerator, signalNGenerator (with KGover = 1) and impSignalNGeneratorFixedP, signalNGeneratorFixedP (with KGover = 0) (any other names welcome !)

cossartque: And in the .dyd each generator could refer to each of these four sets depending on its situation

sources/Outputs/src/Par.cpp
```diff
@@ -131,12 +136,23 @@ Par::writeConstantSets(unsigned int nb_generators, dfl::inputs::Configuration::A
...
-  updateSignalNGenerator(set, activePowerCompensation);
+  updateSignalNGenerator(set, activePowerCompensation, false);
+  ret.push_back(set);
+
+  // Signal N generator with fixed P
+  set = boost::shared_ptr<parameters::ParametersSet>(new parameters::ParametersSet(constants::signalNGeneratorFixedPParId));
+  updateSignalNGenerator(set, activePowerCompensation, true);
   ret.push_back(set);
 
   // Signal N generator with impendance
```
cossartque, line 147: impedance instead of impendance :)

sources/Outputs/src/Par.cpp
```diff
@@ -131,12 +136,23 @@ Par::writeConstantSets(unsigned int nb_generators, dfl::inputs::Configuration::A
...
   set = boost::shared_ptr<parameters::ParametersSet>(new parameters::ParametersSet(constants::impSignalNGeneratorParId));
-  updateSignalNGenerator(set, activePowerCompensation);
+  updateSignalNGenerator(set, activePowerCompensation, false);
+  updateCouplingParameters(set);
+  ret.push_back(set);
+
+  // Signal N generator with impendance with fixed P
```
cossartque, line 153: same

sources/Outputs/src/Par.cpp
PerninGuillaume: That's not mine but I will fix it haha

cossartque (approved):

rosiereflo (changes requested):

rosiereflo (approved):

marcochiaramello (approved):

sources/Outputs/include/Par.h
```diff
@@ -95,9 +95,10 @@ class Par {
    *
    * @param set the parameter set to update
    * @param activePowerCompensation the type of active power compensation
+   * @param fixedP boolean to determines if the set represents a generator with a targetP equal to 0
```
lecourtoisflo, line 98: typo: determine

sources/Outputs/src/Par.cpp
```diff
@@ -194,8 +208,8 @@ Par::writeGenerator(const algo::GeneratorDefinition& def, const std::string& bas
   std::string hashIdStr = std::to_string(hashId);
   auto set = boost::shared_ptr<parameters::ParametersSet>(new parameters::ParametersSet(hashIdStr));
   //  Use the hash id in exported files to prevent use of non-ascii characters
-
-  set->addParameter(helper::buildParameter("generator_KGover", 1.));
+  double value = (DYN::doubleIsZero(def.targetP)) ? 0. : 1.;
```
lecourtoisflo, line 211: These 2 values "0" and "1" could be factorized

tests/algo/TestAlgo.cpp
```diff
@@ -168,25 +168,25 @@ TEST(Generators, base) {
...
+      dfl::algo::GeneratorDefinition("02", dfl::algo::GeneratorDefinition::ModelType::SIGNALN, "2", points, -2, 2, -2, 2, 0),
+      dfl::algo::GeneratorDefinition("05", dfl::algo::GeneratorDefinition::ModelType::SIGNALN, "4", points, -5, 5, -5, 5, 0),
   };
 
   dfl::algo::GeneratorDefinitionAlgorithm::Generators expected_gens_finite = {
-      dfl::algo::GeneratorDefinition("00", dfl::algo::GeneratorDefinition::ModelType::WITH_IMPEDANCE_DIAGRAM_PQ_SIGNALN, "0", points0, 0, 0, 0, 0),
-      dfl::algo::GeneratorDefinition("01", dfl::algo::GeneratorDefinition::ModelType::WITH_IMPEDANCE_DIAGRAM_PQ_SIGNALN, "0", points, -1, 1, -1, 1),
-      dfl::algo::GeneratorDefinition("02", dfl::algo::GeneratorDefinition::ModelType::DIAGRAM_PQ_SIGNALN, "2", points, -2, 2, -2, 2),
-      dfl::algo::GeneratorDefinition("05", dfl::algo::GeneratorDefinition::ModelType::DIAGRAM_PQ_SIGNALN, "4", points, -5, 5, -5, 5),
+      dfl::algo::GeneratorDefinition("00", dfl::algo::GeneratorDefinition::ModelType::WITH_IMPEDANCE_DIAGRAM_PQ_SIGNALN, "0", points0, 0, 0, 0, 0, 0),
```
lecourtoisflo, line 178: Maybe add cases with targetP != 0, unless this parameter has no effect on algorithms

lecourtoisflo (changes requested):

sources/Outputs/src/Par.cpp
PerninGuillaume: How would you do that ?

lecourtoisflo: by defining them as constants of the class

tests/algo/TestAlgo.cpp
PerninGuillaume: The targetP has only an incidence on the Par.cpp and Dyd.cpp, so I don't think it is necessary

sources/Outputs/src/Par.cpp
PerninGuillaume: It is not very clear to me what you mean 
Do you want to add two variables in the class like double zero = 0.; double one = 1.; ?

lecourtoisflo (approved):

rosiereflo:
is there a dependency with a dynawo commit or can we merge this right now?

PR #98, opened 2021-01-08
Hvdc_bug: fixed bug

### LINKED ISSUE #97: Fix the dyd.cpp and the par.cpp for HVDC
When both converters are in the main cc nothing should be written in the dyd.cpp and in the par.cpp for the moment.
It is not the case in the current version.
A test showed that :
- In the .dyd we have a line that implements HvdcPVDangling or HvdcPTanPhiDangling even when both converters are in the main cc, which should not be the case (normally we should not have anything in the .dyd for the moment).
- In the .par we have parameters for an HVDC with both converters in the main cc, which should not be the case.
@marcochiaramello fyi

PerninGuillaume:
closes #97 
based of Dynawo's master

### DIFF
--- a/sources/Inputs/src/NetworkManager.cpp
+++ b/sources/Inputs/src/NetworkManager.cpp
@@ -102,7 +102,7 @@ NetworkManager::buildTree() {
       auto targetP = (*it_g)->getTargetP();
       auto pmin = (*it_g)->getPMin();
       auto pmax = (*it_g)->getPMax();
-      if ((*it_g)->isVoltageRegulationOn() && (DYN::doubleEquals(targetP, pmin) || targetP > pmin) && (DYN::doubleEquals(targetP, pmax) || targetP < pmax)) {
+      if ((*it_g)->isVoltageRegulationOn() && (DYN::doubleEquals(-targetP, pmin) || -targetP > pmin) && (DYN::doubleEquals(-targetP, pmax) || -targetP < pmax)) {
         // We don't use dynamic models for generators with voltage regulation disabled and an active power reference outside the generator's PQ diagram
         nodes_[nodeid]->generators.emplace_back((*it_g)->getID(), (*it_g)->getReactiveCurvesPoints(), (*it_g)->getQMin(), (*it_g)->getQMax(),
                                                 pmin, pmax, targetP);
--- a/sources/Outputs/src/Dyd.cpp
+++ b/sources/Outputs/src/Dyd.cpp
@@ -118,10 +118,18 @@ Dyd::writeHvdcLine(const algo::HvdcLineDefinition& hvdcLine, const std::string& 
   auto model = dynamicdata::BlackBoxModelFactory::newModel(hvdcLine.id);
 
   model->setStaticId(hvdcLine.id);
-  if (hvdcLine.converterType == inputs::HvdcLine::ConverterType::LCC) {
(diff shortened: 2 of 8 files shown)


cossartque (changes requested):
- I propose to put a test (the 12nDynaFlow.iidm)
- Maybe we can also modify the dyd.cpp too. For example, when we choose the lib to use, for the moment we only check the converter's type (LCC or VSC) but we should also check if both converters are in main cc or not. It is not an issue now as you erase the hvdc in the algo.cpp but it will be the case in the future and we will have the same issue when removing the erase line.

PerninGuillaume:
@marcochiaramello I changed the networkManager.cpp to include the minus in front of the targetP since it was needed for the test to pass.

sources/Inputs/src/NetworkManager.cpp
```diff
@@ -102,7 +102,7 @@ NetworkManager::buildTree() {
       auto targetP = (*it_g)->getTargetP();
       auto pmin = (*it_g)->getPMin();
       auto pmax = (*it_g)->getPMax();
-      if ((*it_g)->isVoltageRegulationOn() && (DYN::doubleEquals(targetP, pmin) || targetP > pmin) && (DYN::doubleEquals(targetP, pmax) || targetP < pmax)) {
+      if ((*it_g)->isVoltageRegulationOn() && (DYN::doubleEquals(-targetP, pmin) || -targetP > pmin) && (DYN::doubleEquals(-targetP, pmax) || -targetP < pmax)) {
```
cossartque, line 105: @marcochiaramello is that what you expected here ? Thks!

marcochiaramello: yes!

cossartque (approved):

sources/Outputs/src/Dyd.cpp
```diff
@@ -118,10 +118,18 @@ Dyd::writeHvdcLine(const algo::HvdcLineDefinition& hvdcLine, const std::string&
   auto model = dynamicdata::BlackBoxModelFactory::newModel(hvdcLine.id);
 
   model->setStaticId(hvdcLine.id);
-  if (hvdcLine.converterType == inputs::HvdcLine::ConverterType::LCC) {
-    model->setLib("HvdcPTanPhiDangling");
+  if (hvdcLine.position == algo::HvdcLineDefinition::Position::BOTH_IN_MAIN_COMPONENT) {
+    if (hvdcLine.converterType == inputs::HvdcLine::ConverterType::LCC) {
```
rosiereflo, line 122: If my understandinig is correct we should never be in this case as the hvdcLine is not in the map?

rosiereflo (changes requested):

sources/Outputs/src/Dyd.cpp
cossartque: Yes you are right. But we want to prepare for the future when we will use a dynamic model for this kind of HVDC lines (with both converters in main cc). The " hvdcLines_.erase(it);" line in the Algo.cpp will be removed so we need to have a prepared structure to handle these HVDC lines.

rosiereflo: ok!

rosiereflo (approved):

marcochiaramello (approved):

PR #99, opened 2021-01-11
Update GMock

lecourtoisflo:
FindGMock has been updated in dynawo to make unit tests work but not in child-projects. This was generating a segfault in unit tests when using gmock in debug and gtest in release


lecourtoisflo:
unit tests on centos7 work now

rosiereflo (approved):

Issue #101, opened 2021-01-15
Corrections on Trace following 478 issue in Dynawo


gautierbureau:
https://github.com/dynawo/dynawo/issues/478

https://github.com/dynawo/dynawo/pull/1229

PR #102, opened 2021-01-15
#101 Corrections on Trace following 478 in Dynawo.

gautierbureau:


CMakeLists.txt
```diff
@@ -94,6 +94,8 @@ list(APPEND CMAKE_PREFIX_PATH "${DYNAWO_HOME}/share/cmake")
 list(APPEND CMAKE_MODULE_PATH "${DYNAWO_HOME}/cmake")
 list(APPEND CMAKE_MODULE_PATH "${DYNAWO_HOME}/share/cmake")
 
+set(BOOST_ROOT ${DYNAWO_HOME})
```
lecourtoisflo, line 97: To comply with "standard" way to use "var_ROOT" variables, dynawo_home is only the default value o fBoost_root and log a warning or error if value doesn't match dynawo home

lecourtoisflo (changes requested):

lecourtoisflo:
Also, Boost ROOT should be removed from the README

CMakeLists.txt
gautierbureau: done

gautierbureau:
> Also, Boost ROOT should be removed from the README

done

lecourtoisflo (approved):

PR #103, opened 2021-01-21
Correct compilation and tests

lecourtoisflo:


sources/Outputs/include/Par.h
```diff
@@ -101,8 +101,8 @@ class Par {
    * @param activePowerCompensation the type of active power compensation
    * @param fixedP boolean to determine if the set represents a generator with a targetP equal to 0
    */
-  static boost::shared_ptr<parameters::ParametersSet> updateSignalNGenerator(const std::string& modelId,
-                                     dfl::inputs::Configuration::ActivePowerCompensation activePowerCompensation, bool fixedP);
+  static boost::shared_ptr<parameters::ParametersSet>
+  updateSignalNGenerator(const std::string& modelId, dfl::inputs::Configuration::ActivePowerCompensation activePowerCompensation, bool fixedP);
```
rosiereflo, line 105: I would put two spaces in front of this line

rosiereflo (changes requested):

sources/Outputs/include/Par.h
lecourtoisflo: This is automatic formatting

rosiereflo (approved):

PR #107, opened 2021-02-02
Update following OS 1421

cossartque:
This PR can be merged once the dynawo PR linked to issue 1421 is accepted.

### DIFF
--- a/sources/Outputs/src/Dyd.cpp
+++ b/sources/Outputs/src/Dyd.cpp
@@ -247,7 +247,6 @@ Dyd::writeMacroConnectors() {
 
   connector = dynamicdata::MacroConnectorFactory::newMacroConnector(macroConnectorGenSignalNName_);
   connector->addConnect("generator_N_value", "n_grp_@INDEX@_value");
-  connector->addConnect("generator_alphaSum_value", "alphaSum_grp_@INDEX@_value");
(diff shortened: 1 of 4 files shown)


sources/Outputs/src/Dyd.cpp
```diff
@@ -247,7 +247,6 @@ Dyd::writeMacroConnectors() {
 
   connector = dynamicdata::MacroConnectorFactory::newMacroConnector(macroConnectorGenSignalNName_);
   connector->addConnect("generator_N_value", "n_grp_@INDEX@_value");
-  connector->addConnect("generator_alphaSum_value", "alphaSum_grp_@INDEX@_value");
```
lecourtoisflo, line 250: This has no impacts on MAIN tests ?

lecourtoisflo (changes requested):

sources/Outputs/src/Dyd.cpp
cossartque: It doesn't. This alphSum variable was quite useless. The final result of the simulation is the same. The only change is that in the generator's equations, instead of having (N/alphaSum) we now have N (which has not the same value now of course).

marcochiaramello (approved):

lecourtoisflo (approved):

PR #111, opened 2021-02-05
Remove bus from regulated buses list when the generator's diagram is flat

cossartque:


lecourtoisflo (approved):

sources/Algo/src/Algo.cpp
```diff
@@ -115,6 +115,7 @@ GeneratorDefinitionAlgorithm::operator()(const NodePtr& node) {
     if ((model == GeneratorDefinition::ModelType::DIAGRAM_PQ_SIGNALN || model == GeneratorDefinition::ModelType::REMOTE_DIAGRAM_PQ_SIGNALN ||
          model == GeneratorDefinition::ModelType::PROP_DIAGRAM_PQ_SIGNALN || model == GeneratorDefinition::ModelType::WITH_IMPEDANCE_DIAGRAM_PQ_SIGNALN) &&
         !isDiagramValid(generator)) {
+      busesWithDynamicModel_.erase(generator.regulatedBusId);
```
lecourtoisflo, line 118: in worst case, we have an insert then an erase, which is more time consuming than performing twice the test (model == ...) in switch case

lecourtoisflo (changes requested):

sources/Algo/src/Algo.cpp
```diff
@@ -87,14 +87,23 @@ GeneratorDefinitionAlgorithm::GeneratorDefinitionAlgorithm(Generators& gens, Bus
 void
 GeneratorDefinitionAlgorithm::operator()(const NodePtr& node) {
   auto& node_generators = node->generators;
+
+  auto is_model_diagram = [](GeneratorDefinition::ModelType model, inputs::Generator generator) {
+    return (model == GeneratorDefinition::ModelType::DIAGRAM_PQ_SIGNALN || model == GeneratorDefinition::ModelType::REMOTE_DIAGRAM_PQ_SIGNALN || 
+            model == GeneratorDefinition::ModelType::PROP_DIAGRAM_PQ_SIGNALN || model == GeneratorDefinition::ModelType::WITH_IMPEDANCE_DIAGRAM_PQ_SIGNALN) && 
+            !isDiagramValid(generator);
+ };
```
lecourtoisflo, line 95: alinea is incorrect: please apply formatting

lecourtoisflo (changes requested):
Please do not put names in commit message

sources/Algo/src/Algo.cpp
cossartque: Ok thanks

lecourtoisflo (approved):

rosiereflo (approved):

marcochiaramello (approved):

PR #116, opened 2021-02-12
Update after 1370 OS to call xerces and libxml2 init and terminate me…

rosiereflo:
…thods in main

### DIFF
--- a/sources/main.cpp
+++ b/sources/main.cpp
@@ -95,9 +95,7 @@ main(int argc, char* argv[]) {
       return EXIT_FAILURE;
     }
     DYN::InitXerces xerces;
-#ifdef LANG_CXX11
     DYN::InitLibXml2 libxml2;
-#endif
 
     LOG(info) << MESS(InputsInfo, runtimeConfig.networkFilePath, runtimeConfig.configPath) << LOG_ENDL;
 
--- a/sources/main.cpp
+++ b/sources/main.cpp
@@ -17,12 +17,11 @@
 
 #include <DYNError.h>
 #include <DYNIoDico.h>
+#include <DYNInitXml.h>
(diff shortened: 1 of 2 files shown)


sources/main.cpp
```diff
@@ -93,6 +95,8 @@ main(int argc, char* argv[]) {
       LOG(error) << MESS(NetworkFileNotFound, runtimeConfig.networkFilePath) << LOG_ENDL;
       return EXIT_FAILURE;
     }
+    xmlInitParser();
+    xercesc::XMLPlatformUtils::Initialize();
```
lecourtoisflo, line 99: Doesn't that depend on a USE_XERCES kind of option on dynawo side ?

sources/main.cpp
```diff
@@ -120,6 +126,8 @@ main(int argc, char* argv[]) {
     LOG(info) << " ============================================================ " << LOG_ENDL;
     LOG(info) << MESS(DFLEnded, context.basename(), timerGlobal.elapsed()) << LOG_ENDL;
 
+    xmlCleanupParser();
```
lecourtoisflo, line 129: Factorize

lecourtoisflo (changes requested):

sources/main.cpp
rosiereflo: this is required for all xml parsers beside IIDM

sources/main.cpp
rosiereflo: by that you mean create a method?

lecourtoisflo: Yes, a method that does both lines, since it is used in multiple location

gautierbureau (approved):

lecourtoisflo (approved):

PR #136, opened 2021-04-14
Correct name diagram filename

### LINKED ISSUE #118: Modify the name of the txt file for the diagrams when the name of the generator contains a "/"
In certain cases the name of a generator can contain a "/".
This leads to an error when loading the PQ diagram as the name of the txt file contains this "/".

Maybe we could replace it in the name of the txt file.

lecourtoisflo:
closes #118

### DIFF
--- a/sources/Outputs/src/Constants.cpp
+++ b/sources/Outputs/src/Constants.cpp
@@ -1,0 +1,37 @@
+//
+// Copyright (c) 2020, RTE (http://www.rte-france.com)
+// See AUTHORS.txt
+// All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at http://mozilla.org/MPL/2.0/.
+// SPDX-License-Identifier: MPL-2.0
+//
+
+/**
+ * @file  Constants.cpp
+ *
+ * @brief Dynaflow launcher common to all writers impelmentation file
+ *
+ */
+
+#include "Constants.h"
+
+#include <algorithm>
+
+namespace dfl {
+namespace outputs {
+namespace constants {
+
+std::string
+diagramFilename(const dfl::algo::GeneratorDefinition& generator) {
+  // Remove '/' and '\' characters from id and replace it with '%'
+  auto id = generator.id;
+  std::replace(id.begin(), id.end(), '/', '%');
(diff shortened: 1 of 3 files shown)


scripts/updateMainReference.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python3
```
cossartque, line 1: What is this file for ?

sources/Outputs/src/Constants.cpp
```diff
@@ -0,0 +1,37 @@
+//
+// Copyright (c) 2020, RTE (http://www.rte-france.com)
```
cossartque, line 2: 2021 ?

scripts/updateMainReference.py
rosiereflo: Please add this kind of utility file in a dedicated commit, with the associated doc and comments

sources/Outputs/src/Constants.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+namespace dfl {
+namespace outputs {
+namespace constants {
+
+std::string
+diagramFilename(const dfl::algo::GeneratorDefinition& generator) {
+  // Remove '/' and '\' characters from id and replace it with '%'
+  auto id = generator.id;
+  std::replace(id.begin(), id.end(), '/', '%');
```
rosiereflo, line 30: % can be a special character in Windows if I'm not mistaken (%20 for space for example in html), so I would propose to use '-' or '_'

rosiereflo (changes requested):

sources/Outputs/src/Constants.cpp
marcochiaramello: I'd use '_'

sources/Outputs/src/Constants.cpp
lecourtoisflo: OK

sources/Outputs/src/Constants.cpp
lecourtoisflo: OK

scripts/updateMainReference.py
rosiereflo: I meant a dedicated PR, sorry

cossartque (approved):

rosiereflo (approved):

marcochiaramello (approved):

PR #139, opened 2021-04-15
#138 Add script to update git hook

### LINKED ISSUE #138: Add hooks to comply with dynawo standard
Add git hooks similar to dynawo hooks. That way the checkstyle of the commits will be the same

lecourtoisflo:
closes #138

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # DynaFlow Launcher
 
+Dynawflow Launcher (DFL) depends on the [core dynawo libraries of Dynawo](https://github.com/dynawo/dynawo)
+
 ## Build
 To build DynaFlow launcher, you must first deploy the Dynawo library with c++11 enabled. Then generate the cmake cache in your build directory:
 
@@ -43,3 +45,13 @@ Coverage analysis is based on ctest. To launch coverage, launch from source dire
 `ctest -S cmake/CTestScript.cmake -DDYNAWO_HOME=<DYNAWO_HOME> -DBOOST_ROOT=<DYNAWO_HOME> -VV`
 
 Outputs are gcov and lcov reports
+
+## Developement standard
+### Check style
+The check style, based on Dynawo's, is based on a [clang-format file](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) using [clang-format](https://clang.llvm.org/docs/ClangFormat.html). It is recommended for developpers to apply it before commiting to ensure coherence with future developpements.
+
+### SCM
+git commits have a standard pattern, similar to the one in Dynawo core. This pattern can be checked by hooks. In order to install this hook, the
+developper must run the script:
+
+`./scripts/set_commit_hook.sh`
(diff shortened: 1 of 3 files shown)


scripts/set_commit_hook.sh
```diff
@@ -0,0 +1,105 @@
...
+    fi
+  fi
+}
+
+#################################
+########### Main script #########
+#################################
+
+HERE=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+export DFL_HOME=$HERE/..
```
rosiereflo, line 104: This assumes this scripts was run from the directory of this script if I'm not mistaken

README.md
```diff
@@ -43,3 +45,13 @@ Coverage analysis is based on ctest. To launch coverage, launch from source dire
 `ctest -S cmake/CTestScript.cmake -DDYNAWO_HOME=<DYNAWO_HOME> -DBOOST_ROOT=<DYNAWO_HOME> -VV`
 
 Outputs are gcov and lcov reports
+
+## Developement standard
+### Check style
+The check style, based on Dynawo's, is based on a [clang-format file](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) using [clang-format](https://clang.llvm.org/docs/ClangFormat.html). It is recommended for developpers to apply it before commiting to ensure coherence with future developpements.
+
+### SCM
+git commits have a standard pattern, similar to the one in Dynawo core. This pattern can be checked by hooks. In order to install this hook, the
```
rosiereflo, line 54: not sure it is the best approach to ask the user to do it. It would be better to propose a script similar to envDynawo.sh that provides the compilation service and handles this automatically

README.md
```diff
@@ -1,5 +1,7 @@
 # DynaFlow Launcher
 
+Dynawflow Launcher (DFL) depends on the [core dynawo libraries of Dynawo](https://github.com/dynawo/dynawo)
+
```
rosiereflo, line 4: While we are modifying this file, could you please add a definition of a deployed DYNAWO?

rosiereflo (changes requested):

scripts/set_commit_hook.sh
lecourtoisflo: No HERE is the location of the script file. This assumes the script file is not moved elsewhere before being executed

README.md
lecourtoisflo: DFL has no script for execution and I'm not convinced it should have one. But we can add another issuer for the automation for this step

README.md
lecourtoisflo: I just add "depends on the deployed installation..." ?

README.md
rosiereflo: IMHO It would be coherent with all the tools of the dynawo suite and would ease the compilation of DFL and the test execution for newcomers.
Always good to keep coherency between the projects.

lecourtoisflo: So we create a new issue for that ?

rosiereflo: ok!

rosiereflo (approved):

PR #163, opened 2021-05-18
#161 Correct cmake scripts for cmake 3.12

### LINKED ISSUE #161: adapt CMAKE scripts for older cmake versions
-b option does not exists for older version of cmake (e.g. cmake 3.12). This prevents build.

lecourtoisflo:
closes #161


scripts/envDFL.sh
```diff
@@ -177,9 +177,10 @@ clean() {
 }
 
 cmake_configure() {
+    mkdir -p $DYNAFLOW_LAUNCHER_BUILD_DIR
+    pushd $DYNAFLOW_LAUNCHER_BUILD_DIR
```
rosiereflo, line 181: Could you please redirect the output to void to avoid useless logs?

rosiereflo (changes requested):

rosiereflo (approved):

marcochiaramello (approved):

PR #173, opened 2021-06-24
#172 Add formatting check before commit

### LINKED ISSUE #172: Formatting should be checked before commiting
Formatting should be checked before committing to ensure that coding style is preserved

lecourtoisflo:
closes #172

### DIFF
--- a/scripts/set_commit_hook.sh
+++ b/scripts/set_commit_hook.sh
@@ -78,7 +78,24 @@ fi
 if [[ \"\$whitespace_present\" == \"yes\" || \"\$tab_present\" == \"yes\" ]]; then
   exit 1
 fi
-git diff-index --check --cached HEAD -- ':(exclude)*/reference/*' ':(exclude)*.patch'"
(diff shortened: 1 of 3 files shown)


scripts/set_commit_hook.sh
```diff
@@ -78,7 +78,24 @@ fi
 if [[ \"\$whitespace_present\" == \"yes\" || \"\$tab_present\" == \"yes\" ]]; then
   exit 1
 fi
-git diff-index --check --cached HEAD -- ':(exclude)*/reference/*' ':(exclude)*.patch'"
+git diff-index --check --cached HEAD -- ':(exclude)*/reference/*' ':(exclude)*.patch'
+
+check_format_file() {
+  file=\$1
```
rosiereflo, line 84: I would really prefer to have clang format applied before compilation as this check can be really annoying

rosiereflo (changes requested):

scripts/set_commit_hook.sh
lecourtoisflo: I would do both

rosiereflo (approved):

PR #174, opened 2021-06-25
132 security analysis v2

sheosi:


zamarrenolm (changes requested):

zamarrenolm (approved):

sources/Context.cpp
```diff
@@ -327,6 +397,8 @@ Context::exportOutputs() {
 
   // create output directory
   file::path outputDir(config_.outputDir());
+  outputs::Job::setStartAndDuration(config_.getStartTime(),
```
zamarrenolm, line 400: We should review the definition of start time and duration of simulation in `outputs::Job `class: now they won't be constants, but configurable attributes. They should not be declared static in `Job.h`.

Additionally, for contingency analysis now we will have multiple `Job` instances. Having all the private attributes declared static could lead us into errors later. I think it would be better to have proper objects instead of a `Job` singleton.

sources/Inputs/include/Configuration.h
```diff
@@ -98,6 +98,42 @@ class Configuration {
...
+  float getStartTime() const {
+    return startTime_;
+  }
+
+  /**
+   * @brief Get the Time at which the simulation will end
+   *
+   * @returns the parameter value
+   */
+  float getEndTime() const {
```
zamarrenolm, line 115: `JOB.xsd` in Dynawo uses `stopTime`. Let's keep the same naming convention.

sources/Inputs/include/Configuration.h
```diff
@@ -98,6 +98,42 @@ class Configuration {
     return dsoVoltageLevel_;
   }
 
+  /**
+   * @brief Get the Time at which the simulation will start
+   *
+   * @returns the parameter value
+   */
+  float getStartTime() const {
```
zamarrenolm, line 106: The three times (start, stop, time of event) should be `float` or `double`. Why start and stop are `float` and time of event `double`?.

sources/Inputs/include/Configuration.h
```diff
@@ -125,6 +161,11 @@ class Configuration {
   std::string outputDir_ = boost::filesystem::current_path().generic_string();       ///< Directory for output files
   double dsoVoltageLevel_ = 45.0;                                                    ///< Minimum voltage level of the load to be taken into account
   ActivePowerCompensation activePowerCompensation_ = ActivePowerCompensation::PMAX;  ///< Type of active power compensation
+  float startTime_ = 0;                                                      ///< Moment (in seconds) at which starts the simulation
+  float endTime_ = 100;                                                      ///< Moment (in seconds) at which ends the simulation
```
zamarrenolm, line 165: `stopTime`

sources/Inputs/include/Contingencies.h
```diff
@@ -26,6 +27,37 @@ namespace inputs {
...
+  /**
+   * @brief Explains why an element is considered invalid
+   */
+  enum class ElementInvalidReason {
+    GENERATOR_NOT_FOUND,
+    TWOWINDINGS_TRANFORMER_NOT_FOUND,
+    LINE_NOT_FOUND,
+    BRANCH_NOT_FOUND,
+    SHUNT_COMPENSATOR_NOT_FOUND,
+    DANGLING_LINE_NOT_FOUND,
```
zamarrenolm, line 55: `LOAD_NOT_FOUND` is missing

sources/Inputs/src/Configuration.cpp
```diff
@@ -90,6 +90,10 @@ Configuration::Configuration(const std::string& filepath) {
     helper::updateValue(isAutomaticSlackBusOn_, config, "AutomaticSlackBusOn");
     helper::updateValue(outputDir_, config, "OutputDir");
     helper::updateValue(dsoVoltageLevel_, config, "DsoVoltageLevel");
+    helper::updateValue(startTime_, config, "StartTime");
+    helper::updateValue(endTime_, config, "EndTime");
```
zamarrenolm, line 94: `StopTime`

sources/Outputs/include/Job.h
```diff
@@ -89,9 +90,14 @@ class Job {
    */
   boost::shared_ptr<job::JobEntry> write();
 
+  static void setStartAndDuration(uint32_t startTime, uint32_t durationTime) {
```
zamarrenolm, line 93: Accept parameters as `double` or `float`, whatever we decide. We have to rethink on having these attributes as `static`.

sources/Context.cpp
```diff
@@ -163,162 +201,194 @@ Context::areInMainConnectedComponent(const std::vector<boost::shared_ptr<DYN::Bu
   return result;
```
zamarrenolm, line 201: I think the expression `result = result | nvalid` is wrong/confusing. The operator` | `is a bitwise operator, we should use a logical (boolean) operator, ` || `. Also, the condition does not follow what the name of the function states. The name suggests that a value of true would be returned if all the buses are in the main connected component, thus we should use ` && `. 

Anyway, I think that to check if an element is in the main connected component, it would be enough to check if just one of its end buses is in the main connected component (and there is no need to check the rest) ... we will rewrite and rename the function to reflect this.

zamarrenolm (changes requested):

sources/Context.cpp
```diff
@@ -32,25 +32,54 @@
...
 #include <boost/make_shared.hpp>
+#include <numeric>
 #include <tuple>
 
 namespace file = boost::filesystem;
+using dfl::inputs::Contingencies;
 
 namespace dfl {
+
+void Context::Caches::initCaches(
```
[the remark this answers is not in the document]
sheosi: Roger!

sources/Context.cpp
```diff
@@ -85,6 +114,11 @@ Context::Context(const ContextDef& def, const inputs::Configuration& config) :
     mainConnexIds_.insert(n->id);
```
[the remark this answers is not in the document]
sheosi: Yup, that shouldn't be there...

sources/Context.cpp
```diff
@@ -163,162 +201,194 @@ Context::areInMainConnectedComponent(const std::vector<boost::shared_ptr<DYN::Bu
...
-  for (auto& t : networkManager_.dataInterface()->getNetwork()->getVoltageLevels()) {
-    for (auto& l: t->getLoads()) {
-      if (loadId == l->getID()) {
-        if (!areInMainConnectedComponent(t->getBuses())) {
-          return std::string("both ends of shunt compensator are outside main connected component");
-        }
-        return boost::none; // No problem found
-      }
+  const auto item = caches_.loads.find(loadId);
+  if (item != caches_.loads.find(loadId)) {
```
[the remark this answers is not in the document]
sheosi: Yeah, I slipped on that one

sources/Context.cpp
```diff
@@ -163,162 +201,194 @@ Context::areInMainConnectedComponent(const std::vector<boost::shared_ptr<DYN::Bu
...
+    if (!areInMainConnectedComponent({item->second->getBusInterface()})) {
+      return Contingencies::ElementInvalidReason::NOT_IN_MAIN_CONNECTED_COMPONENT;
     }
+    return boost::none; // No problem found
   }
-  return std::string("not found as shunt compensator");
+  return Contingencies::ElementInvalidReason::SHUNT_COMPENSATOR_NOT_FOUND;
```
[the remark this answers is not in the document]
sheosi: Yup

sources/Context.cpp
```diff
@@ -163,162 +201,194 @@ Context::areInMainConnectedComponent(const std::vector<boost::shared_ptr<DYN::Bu
...
+    case Contingencies::Type::LOAD:
+      return checkLoad(id);
+    case Contingencies::Type::DANGLING_LINE:
+      return checkDanglingLine(id);
+    case Contingencies::Type::HVDC_LINE:
+      return checkHvdcLine(id);
+    case Contingencies::Type::STATIC_VAR_COMPENSATOR:
+      return checkStaticVarCompensator(id);
+    case Contingencies::Type::BUSBAR_SECTION:
+      return boost::none;
```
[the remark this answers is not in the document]
sheosi: Ok

sources/Context.cpp
```diff
@@ -163,162 +201,194 @@ Context::areInMainConnectedComponent(const std::vector<boost::shared_ptr<DYN::Bu
...
-      auto invalid = checkContingencyElement(e->id, e->type);
-      if (invalid) {
-        LOG(warn) << "  Element " << e->id << " (" << e->type << ") not valid, reason: " << invalid.value() << LOG_ENDL;
-        result.push_back(invalid.value());
-      } else {
-        LOG(debug) << "  Element " << e->id << "(" << e->type << ") is valid" << LOG_ENDL;
-      }
+
+  // We'll store here the iterators we don't want
+  std::vector<std::vector<std::shared_ptr<dfl::inputs::Contingencies::ContingencyDefinition>>::iterator> iteratorsToRemove;
```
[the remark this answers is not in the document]
sheosi: Agree, though it is a bit complicated in C++. Functional programming is very verbose and kinda clunky, I event gave it a try, but it was more difficult to read, there's also the problem that we are doing a filter and a fold at the same time (since we are getting the errors).
The good news is that the functional spirit is still preserved, we actually operate on a copy and the whole method is `const`.

sources/Inputs/include/Configuration.h
sheosi: Done

sources/Context.cpp
```diff
@@ -125,17 +159,21 @@ Context::process() {
   onNodeOnMainConnexComponent(algo::ControllerInterfaceDefinitionAlgorithm(hvdcLines_));
   walkNodesMain();
 
-  // Check all contingencies have their elements present in the network
+  // Check all contingencies have their elements present in the network and make
```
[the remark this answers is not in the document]
zamarrenolm: Sorry, it is already in `Context::process`, forget this (but is true that we have to perform it after the caches have been built).

sources/Context.cpp
zamarrenolm: We could leave this change for later. Let's first close a version where we use the Job as it is, a singleton. Just let the start time and duration be changed from the configuration.

sources/Inputs/include/Configuration.h
sheosi: Done

PR #208, opened 2021-09-13
#202 Add default value in case extension with slope is not present

### LINKED ISSUE #202: Manage SVARC components

lecourtoisflo:
closes #202


sources/Outputs/include/Par.h
```diff
@@ -237,6 +237,7 @@ class Par {
   static constexpr double svarcThresholdDown_ = 0.;              ///< time threshold down for SVarC
   static constexpr double svarcThresholdUp_ = 60.;               ///< time threshold up for SVarC
   static constexpr double Sb_ = 100;                             ///< Sb value
+  static constexpr double defaultLambda_ = 0.004;                ///< Default value for lambda parameter
```
cossartque, line 240: For the moment, when the extension is not here, we will not add anything in the .dyd and use the default cpp model.

Issue #209, opened 2021-09-27
Update cpplint configuration


zamarrenolm:
Although the cpplint wrapper considers only source files with `.cpp` extension, the `CPPLINT.cfg` file is missing a line 
```
extensions=cpp,h,hpp,hxx
```
that is present in dynawo. Without this line, modified .cpp files are considered for analysis in the cpplint wrapper, but ignored later in the main cpplint.py script.

zamarrenolm:
If  the `extensions` key is found before the `headers` key in the `CPPLINT.cfg` file, the headers are added to the extensions list. So: 
```
extensions=cpp
headers=h,hpp,hxx
```
Is equivalent to:
```
headers=h,hpp,hxx
extensions=cpp,h,hpp,hxx
```
But to ease future updates/comparisons, CPPLINT.cfg in dynaflow-launcher will be just copied from dynawo.

PR #210, opened 2021-09-27
#209 update cpplint config from dynawo

zamarrenolm:


CPPLINT.cfg
```diff
@@ -1,13 +1,14 @@
...
-root=sources
+extensions=cpp,h,hpp,hxx
+root=dynawo/sources
 linelength=160
 
 # Filter checks
-filter=-legal/copyright
+filter=-build/c++11
 filter=-build/include
+filter=-build/c++11
```
lecourtoisflo, line 10: duplicated filter

CPPLINT.cfg
```diff
@@ -1,13 +1,14 @@
 set noparent
 headers=h,hpp,hxx
-root=sources
+extensions=cpp,h,hpp,hxx
+root=dynawo/sources
```
lecourtoisflo, line 4: dynawo/sources directory doesn't exist in DFL

lecourtoisflo (changes requested):

CPPLINT.cfg
zamarrenolm: Fixed. The cpplint analysis still was performed because the cpplint wrapper is invoked with the proper source directory, and is responsible for building the list of files.

CPPLINT.cfg
zamarrenolm: Fixed. It is also duplicated in dynawo.

CPPLINT.cfg
```diff
@@ -1,13 +1,13 @@
...
 # Filter checks
-filter=-legal/copyright
+filter=-build/c++11
 filter=-build/include
 filter=-build/include_what_you_use
 filter=-build/include_order
-filter=-build/c++11
 filter=-runtime/references
-filter=-runtime/string
+filter=-runtime/int
```
lecourtoisflo, line 13: Why did you add this one ? verboten C basic types should not be used

lecourtoisflo (changes requested):

CPPLINT.cfg
zamarrenolm: It comes from the configuration used in dynawo. I just copied the current CPPLINT.cfg found in master branch of dynawo. Should this filter be removed here and also in dynawo ?

lecourtoisflo: In dynawo, it shouldn't. But in DFL, maybe it should because we have less constraints. @gautierbureau what do you think ?

lecourtoisflo: After looking at the definitions of the filters, runtime/string and runtime/int should be present in cpplint of DFL

zamarrenolm: Done, the line for runtime/string was recovered

lecourtoisflo: No it is the opposide, the filters should be checked. Your lines removed the check

zamarrenolm: Sorry for the confusion. I updated CPPLINT.cfg so both filters are not discarded.

lecourtoisflo (approved):

marcochiaramello (approved):

PR #221, opened 2021-10-04
#220 Fix HVDC model when VSC are connected by switch

### LINKED ISSUE #220: Fix case when two VSC are connected to nodes connected by a switch
When two VSC are connected to and regulate two buses that are connected by a retained switch, the PQProp model with a SignalNQ model should be used (same logic as for the generators).

@marcochiaramello fyi

lecourtoisflo:
closes #220


lecourtoisflo:
Validated with network remote control test

cossartque (approved):

PR #230, opened 2021-11-03
#229 Slack node not fictitious

### LINKED ISSUE #229: Do not consider fictitious buses in slack node selection
Dynawo may create fictitious buses in some situations. For example when converting three-winding transformers. These buses must not be considered for slack node selection.

zamarrenolm:
Closes #229.
Requires Dynawo explicit flag for fictitious bus component interface: dynawo/dynawo#1954

### DIFF
--- a/sources/Inputs/src/Node.cpp
+++ b/sources/Inputs/src/Node.cpp
@@ -21,26 +21,18 @@ namespace dfl {
 namespace inputs {
 
 std::shared_ptr<Node>
-Node::build(const NodeId& id, const std::shared_ptr<VoltageLevel>& vl, double nominalVoltage, const std::vector<Shunt>& shunts) {
-  bool fictitious = false;
-  auto ret = std::shared_ptr<Node>(new Node(id, vl, nominalVoltage, fictitious, shunts));
+Node::build(const NodeId& id, const std::shared_ptr<VoltageLevel>& vl, double nominalVoltage, const std::vector<Shunt>& shunts, bool fictitious) {
+  auto ret = std::shared_ptr<Node>(new Node(id, vl, nominalVoltage, shunts, fictitious));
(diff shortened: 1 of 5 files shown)


zamarrenolm:
Rebased on current master and labelled as ready for review after dependant Dynawo change has been merged (dynawo/dynawo#1954).

sources/Inputs/src/Node.cpp
```diff
@@ -22,15 +22,24 @@ namespace inputs {
...
-  auto ret = std::shared_ptr<Node>(new Node(id, vl, nominalVoltage, shunts));
+  bool fictitious = false;
+  auto ret = std::shared_ptr<Node>(new Node(id, vl, nominalVoltage, fictitious, shunts));
   vl->nodes.push_back(ret);
   return ret;
 }
 
-Node::Node(const NodeId& idNode, const std::shared_ptr<VoltageLevel> vl, double nominalVoltageNode, const std::vector<Shunt>& shunts) :
+std::shared_ptr<Node>
+Node::build(const NodeId& id, const std::shared_ptr<VoltageLevel>& vl, double nominalVoltage, bool fictitious, const std::vector<Shunt>& shunts) {
```
lecourtoisflo, line 32: It is better in that case to put fictitious as the last argument of the list and have the default value "false" than creating another builder

lecourtoisflo (changes requested):

sources/Inputs/src/Node.cpp
zamarrenolm: Ok. I will also change the private constructor to keep the same order in the arguments.

zamarrenolm: Fixed

lecourtoisflo (approved):

gautierbureau (approved):

PR #234, opened 2021-11-09
#233 Copy all Static Var Compensators and generators to inputs

### LINKED ISSUE #233: Store all SVarCs and generators in inputs even if they will receive default model
The agreed validation scheme of contingencies in the security analysis simulation requires that all network elements that could be reference in a contingency are present in the DFL inputs. 

Currently, Static Var Compensators and generators that will not receive a specific dynamic model in the DYD are not copied in the DFL inputs data structures.

These elements should be stored in the DFL inputs, with a flag indicating that no specific model will be assigned (a default model will be assigned by Dynawo).

zamarrenolm:
Closes #233


sources/Inputs/src/NetworkManager.cpp
```diff
@@ -151,22 +151,21 @@ NetworkManager::buildTree() {
     }
 
     const auto& svarcs = networkVL->getStaticVarCompensators();
+    bool useDefaultModel = false;
```
lecourtoisflo, line 154: init of useDefaultModel shall be done at each loop

sources/Inputs/include/Behaviours.h
```diff
@@ -178,9 +178,10 @@ struct StaticVarCompensator {
    * @param USetPointMax the high voltage set point of the SVarC
    * @param b0 the initial susceptance value  of the SVarC
    * @param slope the slope (kV/MVar) of the SVarC
+   * @param useDefaultModel true if default model is going to be used for the SVarC
```
lecourtoisflo, line 181: From DFL point of view, it is weird to have the information "use default model" connected to the action "do nothing", even if is relevant from dynawo point of view. Replace default model by something like "exportSvarcModel"

lecourtoisflo (changes requested):

sources/Inputs/include/Behaviours.h
zamarrenolm: What about `defineSvarcModel` ?

lecourtoisflo: OK

zamarrenolm: Done. I will use the same semantics for the new flag that I'm adding for the generators. I plan to include it in this  PR.

sources/Inputs/src/NetworkManager.cpp
zamarrenolm: Fixed. Added more elements to the PAR and DYD unit tests to verify.

sources/Inputs/src/NetworkManager.cpp
```diff
@@ -152,21 +152,20 @@ NetworkManager::buildTree() {
 
     const auto& svarcs = networkVL->getStaticVarCompensators();
     for (const auto& svarc : svarcs) {
+      bool defineSvarcModel = true;
       if (!svarc->getInitialConnected()) {
-        continue;
-      }
-      if (!svarc->hasStandbyAutomaton()) {
+        defineSvarcModel = false;
```
lecourtoisflo, line 157: Add a warning message like the others for compliance in this specific case

lecourtoisflo (changes requested):

sources/Inputs/src/NetworkManager.cpp
zamarrenolm: Message added

lecourtoisflo (approved):

lecourtoisflo (approved):

zamarrenolm:
Commits squashed

lecourtoisflo:
> Commits squashed

You should only squash commits before the PR creation. Then each additionnal commit should independant. Squashing should be done only when additionnal commits are too small to be relevant alone, and the first commit is not squash with the others

zamarrenolm:
> > Commits squashed
> 
> You should only squash commits before the PR creation. Then each additionnal commit should independant. Squashing should be done only when additionnal commits are too small to be relevant alone, and the first commit is not squash with the others

Sorry, I thought the common practice was to squash all commits in the PR after it has been approved, so the final merge would contain only one set of changes. I was trying to leave the PR _ready for merge_.

lecourtoisflo:
> > > Commits squashed
> > 
> > 
> > You should only squash commits before the PR creation. Then each additionnal commit should independant. Squashing should be done only when additionnal commits are too small to be relevant alone, and the first commit is not squash with the others
> 
> Sorry, I thought the common practice was to squash all commits in the PR after it has been approved, so the final merge would contain only one set of changes. I was trying to leave the PR _ready for merge_.

The practice for dynawo is we have one commit for the PR creation and then we add new commit to take remarks into account, in a minimal number of commit for readability. Your PR on SA was an exception because there were a lot of commits

etc/DFLMessages_en_GB.dic
```diff
@@ -16,7 +16,9 @@ UnsupportedCountName          =     Unsupported count name %1% for setting set
 UnsupportedDataTypeReference  =     Unsupported data type %1% for reference %2% in setting file
 UnsupportedOrigDataReference  =     Unsupported data origin %1% for reference %2% in setting file
 UnsupportedParameterDataType  =     Unsupported set parameter type %1% in setting file
-SVarCIIDMExtensionNotFound    =     IIDM extension %1% not found for static var compensator %2% : it will be ignored
+SVarCInitiallyDisconnected    =     Static var compensator %1 is initially disconnected : default model will be used
```
marcochiaramello, line 19: missing "%" should be "%1%"

zamarrenolm: Message fixed

zamarrenolm:
Solved conflicts with master branch

marcochiaramello (approved):

zamarrenolm:
rebased on current master

Issue #15, opened 2020-10-07
Continuous integration


rosiereflo:
Implement with github actions to be compliant with dynawo workflow

PR #270, opened 2022-01-17
#15 Fix fail in main_sa unitttest because of missing compilation flag.

gautierbureau:


gautierbureau:
The fail is now due the a real fail on the master related to dynawo and not my correction that is correct !

PR #277, opened 2022-01-18
#258 avoiding writing useless informations in par and dyd when a comp…

### LINKED ISSUE #258: Avoiding writing of useless informations in par and dyd when an object does not exists
In case a component it is not present in the input file (e.g. SvarCs, loads, ...) no information should be written in dynawo input files

marcochiaramello:
…onent definition does not exists
Closes #258


sources/Outputs/include/Dyd.h
```diff
@@ -191,27 +191,33 @@ class Dyd {
    *
    * Create Model signal N model
    *
+   * @param def the dyd definition
+   *
    * @returns black box models
    */
-  static std::vector<boost::shared_ptr<dynamicdata::BlackBoxModel>> writeConstantsModel();
+  static std::vector<boost::shared_ptr<dynamicdata::BlackBoxModel>> writeConstantsModel(const DydDefinition& def_);
```
rosiereflo, line 198: no "_" in parameter names

tests/outputs/TestPar.cpp
```diff
@@ -244,7 +244,7 @@ TEST(TestPar, writeStaticVarCompensator) {
   outputPath.append(filename);
   dfl::inputs::Configuration::ActivePowerCompensation activePowerCompensation(dfl::inputs::Configuration::ActivePowerCompensation::P);
   dfl::outputs::Par parWriter(dfl::outputs::Par::ParDefinition(basename, dirname, outputPath.generic_string(), {}, {}, activePowerCompensation, {},
-                                                               manager, {}, {}, {}, svarcs));
+                                                               manager, {}, {}, {}, svarcs, {}));
```
rosiereflo, line 247: Maybe adding a test with no load to test that the connection is not dumped? (same goes for dyd)

rosiereflo (changes requested):

sources/Outputs/include/Dyd.h
marcochiaramello: ok, thx

tests/outputs/TestPar.cpp
marcochiaramello: I'd rather add a test on loads as missing, for the other tests loads are already missing

rosiereflo (approved):

PR #293, opened 2022-01-21
#280 remove results folder before running tests again

### LINKED ISSUE #280: Delete results folder before launching tests
If old results are present they can be used for comparison and say the tests are ok whereas in fact they are not... a clean environment like in the CI would detect those issues. Something like `find tests -type d -name "results" -exec rm -rf {} \;` would solve the issue

marcochiaramello:
Closes #280


scripts/envDFL.sh
```diff
@@ -265,6 +265,8 @@ cmake_build() {
 }
 
 cmake_tests() {
+    pushd $DYNAFLOW_LAUNCHER_HOME > /dev/null
```
gautierbureau, line 268: you need a popd after the find then

gautierbureau (changes requested):

scripts/envDFL.sh
marcochiaramello: done , thx

gautierbureau (approved):

gautierbureau:
Thinking about it I think we should maybe have a name less generic than "results" because if somebody has a folder with this name in its folder it will delete it... Maybe resultsTestsTmp to make sure nobody would use the name ?

marcochiaramello:
> Thinking about it I think we should maybe have a name less generic than "results" because if somebody has a folder with this name in its folder it will delete it... Maybe resultsTestsTmp to make sure nobody would use the name ?

I quite agree with the proposal, I'll more results into resultsTestsTmp folder everywhere and adapt the fix

gautierbureau (approved):

PR #324, opened 2022-02-24
#323 Have one switchOffSignal/running for each side of an HVDC (OS 2191)

cossartque:
This requires OS 2191 and dynawo-rte 432 to be merged first


cossartque:
The tests fail but that's because dynawo branch is not merged yet

adriguir (approved):

PR #325, opened 2022-03-02
#70 add precision to config file

### LINKED ISSUE #70: Set precision attribute
Dynawo's core allows to set a certain precision to the simulation. It would be nice and interesting to have a constant in the DynaFlow launcher code to set this precision. 

It could be discussed later on if it needs to be modifiable through a platform calling DFL.

menezesian:
Closes #70


sources/Outputs/include/Job.h
```diff
@@ -104,8 +106,8 @@ class Job {
   * @param networkFileEntry path to the input network file
   * @param outputDir the output directory
   */
-  static void exportJob(const boost::shared_ptr<job::JobEntry>& jobEntry, const boost::filesystem::path& networkFileEntry,
```
gautierbureau, line 107: why change the signature of the method ?

gautierbureau (changes requested):

marcochiaramello (changes requested):

rosiereflo (approved):
Nothing to add to others reviewer comments

sources/Outputs/include/Job.h
menezesian: Because I had to access a member attribute inside this method to ensure that we would output the new precision only if it was set by the user. Otherwise Dynaflow would use the same value as Dynawo and there would be no need to write it again

rosiereflo: Please pass this information as a parameter of the method

sources/Context.cpp
```diff
@@ -211,12 +211,12 @@ Context::exportOutputJob() {
   switch (def_.simulationKind) {
   case SimulationKind::SECURITY_ANALYSIS:
     // For security analysis always export the main jobs file, as dynawo-algorithms will need it
-    outputs::Job::exportJob(jobEntry_, absolute(def_.networkFilepath), config_.outputDir());
+    jobWriter.exportJob(jobEntry_, absolute(def_.networkFilepath), config_.outputDir());
```
rosiereflo, line 214: as it is static please call through the class name

sources/Outputs/include/Job.h
```diff
@@ -72,28 +67,19 @@ class Job {
...
         contingencyId(contingencyId),
-        baseFilename(baseFilename) {
-      initFromConfig(config);
-    }
+        baseFilename(baseFilename),
+        precision_(config.getPrecision()) {}
 
     std::string filename;                        ///< filename of the job output file
     std::string dynawoLogLevel;                  ///< Dynawo log level, in string representation
-    Seconds startTime = Seconds(0);              ///< The start time of the simulation
```
rosiereflo, line 82: Why removing the default values for those parameters?

rosiereflo (changes requested):

sources/Outputs/include/Job.h
```diff
@@ -40,24 +40,19 @@ class Job {
    public:
     using Seconds = std::chrono::seconds;  ///< Alias for seconds
 
-    /**
-     * @brief Constructor
-     *
-     * @param filename output filename
-     * @param lvl dynawo log level
-     */
-    JobDefinition(const std::string& filename, const std::string& lvl) : filename(filename), dynawoLogLevel(lvl) {}
```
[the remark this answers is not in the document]
marcochiaramello: this is because this method was only used in tests

sources/Outputs/include/Job.h
menezesian: Because they are already defined for the Configuration. Since all the constructors have a Config as parameter, we would be defining default values in two different places

sources/Context.cpp
menezesian: Done

sources/Outputs/src/Job.cpp
```diff
@@ -227,6 +230,7 @@ Job::exportJob(const boost::shared_ptr<job::JobEntry>& jobEntry, const boost::fi
   auto simu = jobEntry->getSimulationEntry();
   attrs.add("startTime", simu->getStartTime());
   attrs.add("stopTime", simu->getStopTime());
+  attrs.add("precision", simu->getPrecision());
```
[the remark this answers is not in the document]
menezesian: In the writeOutputs method ?

sources/Outputs/include/Job.h
menezesian: It's back to static

marcochiaramello (approved):

rosiereflo (approved):

gautierbureau (approved):

PR #346, opened 2022-03-22
#345 when a load is fictitous, associate a network model

### LINKED ISSUE #345: When a load is fictitious, a network model should be associated

marcochiaramello:
Closes #345

### DIFF
--- a/tests/algo/TestAlgo.cpp
+++ b/tests/algo/TestAlgo.cpp
@@ -469,25 +469,22 @@ TEST(Generators, SwitchConnexity) {
 TEST(Loads, base) {
   auto vl = std::make_shared<dfl::inputs::VoltageLevel>("VL");
   using dfl::algo::LoadDefinition;
-  std::vector<std::shared_ptr<dfl::inputs::Node>> nodes{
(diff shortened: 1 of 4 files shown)


tests/algo/TestAlgo.cpp
```diff
@@ -469,25 +469,22 @@ TEST(Generators, SwitchConnexity) {
...
-      LoadDefinition("05", LoadDefinition::ModelType::NETWORK, "4"),
-  };
+      LoadDefinition("01", LoadDefinition::ModelType::LOADRESTORATIVEWITHLIMITS, "0"), LoadDefinition("02", LoadDefinition::ModelType::NETWORK, "0"),
+      LoadDefinition("03", LoadDefinition::ModelType::NETWORK, "2"), LoadDefinition("04", LoadDefinition::ModelType::NETWORK, "4")};
 
-  nodes[0]->loads.emplace_back("00");
-  nodes[0]->loads.emplace_back("01");
+  nodes[0]->loads.emplace_back("00", false);
+  nodes[0]->loads.emplace_back("01", false);
+  nodes[0]->loads.emplace_back("02", true);
```
rosiereflo, line 483: No test that this load has a network type?

marcochiaramello: it is tested few line below, where we check that id and model found is equal to what expected (line 500, 501)

PR #354, opened 2022-03-30
#338 do not run the base case in SA as we assume the scenario input iidm

### LINKED ISSUE #338: Base case should not be run in SA
the run of the base case should be done previously than running the SA. it is up to the user to give the output iidm from base case as input for the SA. Tests should be updated accordingly. In addition to that, results should be systematically written

rosiereflo:
is already in steady-state
closes #338


marcochiaramello:
we should also dump the aggregated results systematically, can you add this? I am also thinking the powsybl would need only the aggregated results and constraints and not the entire output iidm. This last could be dumped only in debug mode

rosiereflo:
> we should also dump the aggregated results systematically, can you add this? I am also thinking the powsybl would need only the aggregated results and constraints and not the entire output iidm. This last could be dumped only in debug mode

done

marcochiaramello (approved):

PR #362, opened 2022-04-08
#358 Simplify environment to use a different dynawo home at runtime and

### LINKED ISSUE #358: [Compilation] DYNAFLOW_LAUNCHER_EXTERNAL* environment variables  should be used during installation to get a standalone folder
We should also take a look at how they are used to ease their usage

rosiereflo:
use this information during the installation
closes #358


gautierbureau:
For me all this mechanism is not really usefull and we should get rid of it completly. We can merge the correction to have it working before removing it. I don't see the point of having other env variables to complicate the environment whereas we could just use DYNAWO_HOME to do exactly the same behavior

rosiereflo:
> For me all this mechanism is not really usefull and we should get rid of it completly. We can merge the correction to have it working before removing it. I don't see the point of having other env variables to complicate the environment whereas we could just use DYNAWO_HOME to do exactly the same behavior

Yes you're right, the DYNAWO_IIDM_EXTENSION would still be needed though

rosiereflo:
Let me try your idea of directly using DYNAWO_HOME and sees if it works

gautierbureau (approved):

Issue #396, opened 2022-05-19
Output slack node in a separate file


gautierbureau:
currently output is in log

rosiereflo:
This needs to be added in v1.3 branch

PR #399, opened 2022-05-19
#396 Ouput slack node in separate file.

gautierbureau:

### DIFF
--- a/sources/Context.cpp
+++ b/sources/Context.cpp
@@ -38,6 +38,7 @@
 #include <algorithm>
 #include <boost/filesystem.hpp>
 #include <boost/make_shared.hpp>
+#include <boost/property_tree/json_parser.hpp>
 #include <tuple>
 
 namespace file = boost::filesystem;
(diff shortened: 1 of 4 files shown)


gautierbureau:
I updated with a json file

gautierbureau:
Linked powsybl PR : https://github.com/powsybl/powsybl-dynawo/pull/106

sources/Context.cpp
```diff
@@ -344,6 +345,37 @@ Context::executeSecurityAnalysis() {
   saLauncher->writeResults();
 }
 
+void
+Context::exportResults(bool isOk) {
+  switch (def_.simulationKind) {
+  case SimulationKind::STEADY_STATE_CALCULATION: {
+    boost::property_tree::ptree resultsTree, componentResultsTree, componentResultsChild;
+    resultsTree.put("version", "1.2");
```
rosiereflo, line 353: Is this the powsybl version? If so we are using 1.4

sources/Context.h
```diff
@@ -106,6 +106,15 @@ class Context {
    */
   void exportOutputs();
 
+  /**
+   * @brief Export results file
+   *
+   * @param isOk the simulation was ok
```
rosiereflo, line 112: Maybe we could call this parameter "simulationOk"?

rosiereflo:
There is also 1 code smell to fix

sources/Context.cpp
gautierbureau: No this is the version of the LodFlowResult format. See https://github.com/powsybl/powsybl-core/blob/main/loadflow/loadflow-api/src/test/resources/LoadFlowResultVersion12.json

sources/Context.cpp
```diff
@@ -344,6 +345,37 @@ Context::executeSecurityAnalysis() {
...
+    resultsTree.put("isOK", isOk);
```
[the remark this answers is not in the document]
gautierbureau: it is mandatory in the result format

sources/Context.h
gautierbureau: we can but the name of the parameter in the json needs to be isOk

gautierbureau:
> There is also 1 code smell to fix

If fixed it and solved other issues

CMakeLists.txt
```diff
@@ -64,7 +64,7 @@ set(FORCE_CXX11_ABI OFF CACHE BOOL "Choose either ON or OFF.")
 
 option(DYNAFLOW_LAUNCHER_USE_ADDITIONAL_SOURCEFILE "Use the additional source file" ON)
 
-option(DYNAFLOW_LAUNCHER_SHARED_LIB "Compile shared library" OFF)
+option(DYNAFLOW_LAUNCHER_SHARED_LIB "Compile shared library" ON)
```
rosiereflo, line 67: Not related, should be done in another PR

rosiereflo (changes requested):

rosiereflo (approved):

Issue #356, opened 2022-03-30
[Outputs] Minimize the number of outputs of SA


rosiereflo:
platform (PowSybl)  needs only the aggregated results and some output files (contraints, lost equipments...) and not the entire output iidm. This last could be dumped only in debug mode

PR #400, opened 2022-05-19
#396 Ouput slack node in separate file.

gautierbureau:

- [x] if this commit targets a release branch: the corresponding milestone was added in the ticket and in this PR

### DIFF
--- a/sources/CMakeLists.txt
+++ b/sources/CMakeLists.txt
@@ -41,6 +41,8 @@ target_link_libraries(DynaFlowLauncher
   DynawoAlgorithms::dynawo_algorithms_Launcher
 
   LibXml2::LibXml2
+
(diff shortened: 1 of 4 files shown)


rosiereflo:
I think we should properly brainstorm about DFL outputs and what we want to put in there (see https://github.com/dynawo/dynaflow-launcher/issues/356)

gautierbureau:
> I think we should properly brainstorm about DFL outputs and what we want to put in there (see #356)

We can. This PR is related to this one in powsybl-dynawo : https://github.com/powsybl/powsybl-dynawo/pull/106 . We need to complete the LoadFlowResultImpl to comply with API.

gautierbureau:
I will do another pr with the proper commits when the master version is accepted

PR #408, opened 2022-06-02
407 add timestep choice in config

marcochiaramello:

- [x] main documentation was updated (update of input/output file)

### DIFF
--- a/documentation/functionalDoc/functionalDoc.tex
+++ b/documentation/functionalDoc/functionalDoc.tex
@@ -232,10 +232,14 @@ The conditions and models used are given below.
 \section{Solver}
 
 The solver used is the Simplified Solver (SolverSIM) from \Dynawo.
-The Maximum time step value is 10s.
+The Maximum time step value is by default 10s. This value can be modified in the configuration file (see section \ref{Dynaflow_Launcher_Configuration_Configuration_File}).
+Notice that the maximum step value should be reduced only when the simulation is including
+Special Protection Scheme automatons with a timescale smaller than 10 seconds: this would allow one to properly
+take into account the SPS behaviour, avoiding any artificial synchronization effect. When modifying the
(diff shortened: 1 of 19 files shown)


sources/Outputs/src/Solver.cpp
```diff
@@ -0,0 +1,74 @@
...
+// See AUTHORS.txt
+// All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at http://mozilla.org/MPL/2.0/.
+// SPDX-License-Identifier: MPL-2.0
+//
+
+/**
+ * @file  Par.cpp
```
rosiereflo, line 12: Wrong file name

documentation/functionalDoc/functionalDoc.tex
```diff
@@ -232,10 +232,13 @@ \subsection{Static Var Compensators}
 \section{Solver}
 
 The solver used is the Simplified Solver (SolverSIM) from \Dynawo.
-The Maximum time step value is 10s.
+The Maximum time step value is by default 10s. This value can be modified in the configuration file (see section \ref{Dynaflow_Launcher_Configuration_Configuration_File}).
```
rosiereflo, line 235: Maybe we should specify that the value provided should be above 1 (hMin)

rosiereflo (changes requested):

sources/Outputs/src/Solver.cpp
marcochiaramello: thx

documentation/functionalDoc/functionalDoc.tex
marcochiaramello: done, thx

PR #499, opened 2022-11-24
#497 factorize the method to select generators parId

### LINKED ISSUE #497: Factorize the method to select parameters id for generator
Today the information is duplicated in DydGenerator.cpp and ParGenerator.cpp and (of course) there is a mismatch....

rosiereflo:
closes #497

### DIFF
--- a/sources/Outputs/src/ParGenerator.cpp
+++ b/sources/Outputs/src/ParGenerator.cpp
@@ -25,9 +25,9 @@ ParGenerator::write(boost::shared_ptr<parameters::ParametersSetCollection>& para
       continue;
     }
     boost::shared_ptr<parameters::ParametersSet> paramSet;
-    if (!generator.isUsingDiagram()) {
-      if (!paramSetCollection->hasParametersSet(getGeneratorParameterSetId(generator.model, DYN::doubleIsZero(generator.targetP)))) {
-        paramSet = writeConstantGeneratorsSets(activePowerCompensation, generator.model, generator.targetP, startingPointMode);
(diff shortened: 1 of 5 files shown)


sources/Outputs/src/ParGenerator.cpp
```diff
@@ -25,9 +25,9 @@ ParGenerator::write(boost::shared_ptr<parameters::ParametersSetCollection>& para
       continue;
     }
     boost::shared_ptr<parameters::ParametersSet> paramSet;
-    if (!generator.isUsingDiagram()) {
-      if (!paramSetCollection->hasParametersSet(getGeneratorParameterSetId(generator.model, DYN::doubleIsZero(generator.targetP)))) {
-        paramSet = writeConstantGeneratorsSets(activePowerCompensation, generator.model, generator.targetP, startingPointMode);
+    if (helper::generatorSharesParId(generator)) {
```
marcochiaramello, line 28: from the point of view of behaviour, this seems correct to me. What I do not like, is that we loose the info we are using infinite diagrams here

marcochiaramello (changes requested):

sources/Outputs/src/ParGenerator.cpp
rosiereflo: The issue is that the information should be held at one and one place only: it was before given in DydGenerator, ParGenerator and this test on using or not a diagramp was a 3rd assumption on what was returned in ParGenerator...

marcochiaramello (approved):

PR #503, opened 2022-11-25
#502 fix version number in produced results json file

marcochiaramello:

### DIFF
--- a/sources/Context.cpp
+++ b/sources/Context.cpp
@@ -28,6 +28,7 @@
 #include "Par.h"
 #include "ParEvent.h"
 #include "Solver.h"
+#include "version.h"
 
 #include <DYNMPIContext.h>
 #include <DYNMultipleJobsFactory.h>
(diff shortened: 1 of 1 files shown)


sources/Context.cpp
```diff
@@ -365,7 +356,7 @@ Context::exportResults(bool simulationOk) {
     boost::property_tree::ptree resultsTree;
     boost::property_tree::ptree componentResultsTree;
     boost::property_tree::ptree componentResultsChild;
-    resultsTree.put("version", "1.2");
+    resultsTree.put("version", DYNAFLOW_LAUNCHER_VERSION_STRING);
```
rosiereflo, line 359: This is not Dynawo version but the version of the result file if I'm not mistaken @gautierbureau do you confirm?

gautierbureau: Yes exactly. And for info they are now at 1.3 https://github.com/powsybl/powsybl-core/commit/cc3312890d0413f4a344ed37bcb68f458dc14944 but it seems the new attribute is optional so we can keep 1.2 for the moment

marcochiaramello: ok, sorry, my bad then!

PR #534, opened 2023-01-06
489 having configuration json analysis case insensitive

barondim:


tests/inputs/TestConfig.cpp
```diff
@@ -24,6 +24,19 @@ TEST(Config, Incorrect) {
                       dfl::KeyError_t::InvalidActivePowerCompensation);
 }
 
+TEST(Config, CaseInsensitive) {
```
rosiereflo, line 27: Maybe we should add a test with a boolean badly written?

rosiereflo (approved):

Issue #535, opened 2023-01-06
It shouldn't be possible to execute the option build-tests-coverage in Release


barondim:
Normally, the option `build-tests-coverage` should force the build in Debug like in the other projects. Currently, it's possible to launch the coverage in Release and it returns tests which are KO.

PR #552, opened 2023-01-18
#535 TUs during coverage now pass when the project is in Release beca…

barondim:
…use we now link with .so in ./build/coverage

### DIFF
--- a/scripts/envDFL.sh
+++ b/scripts/envDFL.sh
@@ -210,6 +210,7 @@ env_var_sanity_check() {
 }
 
 set_environment() {
+    MODE=$1
     env_var_sanity_check
     export_var_env DYNAFLOW_LAUNCHER_HOME=UNDEFINED
     export_var_env DYNAFLOW_LAUNCHER_BUILD_TYPE=UNDEFINED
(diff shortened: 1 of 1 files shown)


barondim:
- [ ] We need to discuss if this is the right solution before merging

scripts/envDFL.sh
```diff
@@ -255,8 +256,10 @@ set_environment() {
     export DYNAWO_LIBIIDM_EXTENSIONS=$DYNAWO_INSTALL_DIR/lib
     export DYNAWO_RESOURCES_DIR=$DYNAFLOW_LAUNCHER_INSTALL_DIR/share:$DYNAWO_INSTALL_DIR/share/xsd
 
-    ld_library_path_prepend $DYNAFLOW_LAUNCHER_INSTALL_DIR/lib64 # For local DFL libraries, used only at runtime in case we compile in shared
-    ld_library_path_prepend $DYNAFLOW_LAUNCHER_INSTALL_DIR/lib # For local DFL libraries, used only at runtime in case we compile in shared
+    if [ $MODE -ne 2 ]; then
+        ld_library_path_prepend $DYNAFLOW_LAUNCHER_INSTALL_DIR/lib64 # For local DFL libraries, used only at runtime in case we compile in shared
```
rosiereflo, line 260: Could you please explain the rational behind this fix?

barondim: The fix applies a different mode if we run the tests or the coverage.

It fixed the problem about the user being able to execute the option `build-tests-coverage` using a Release build. Before, the coverage used the user build to run. Now, the coverage uses the coverage build in `/build/coverage` which is always built in Debug. To do this, @gautierbureau removed the `$DYNAFLOW_LAUNCHER_INSTALL_DIR/lib64` and `$DYNAFLOW_LAUNCHER_INSTALL_DIR/lib` which was previously added to the `LD_LIBRARY_PATH`. This way, the coverage will only link with the .so files of the coverage which are built in Debug type. Previously, it searched for the .so files of the user build in the `LD_LIBRARY_PATH` which could be built in Release type... which leads to the error.

rosiereflo (approved):

Issue #575, opened 2023-02-06
Update NRT references after timeline times are filled out with dynawo precision


barondim:
Consequence of https://github.com/dynawo/dynawo/issues/2762

PR #576, opened 2023-02-06
#575 NRT update

barondim:

### DIFF
--- a/scripts/checkContingencies.py
+++ b/scripts/checkContingencies.py
@@ -8,7 +8,8 @@ import re
 import xml.etree.ElementTree as ET
 
 ### Regexes ####################################################################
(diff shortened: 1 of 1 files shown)


scripts/checkContingencies.py
```diff
@@ -8,8 +8,7 @@
 import xml.etree.ElementTree as ET
 
 ### Regexes ####################################################################
-contingency_event_time = 10
-timeline_log_regex = re.compile("^" + str(contingency_event_time) + "\s\|\s([\w-]+)\s\|\s\w+\s[:\s]*[\w\s]+$", re.MULTILINE)
+timeline_log_regex = re.compile("^10\.000\s\|\s([\w-]+)\s\|\s\w+\s[:\s]*[\w\s]+$", re.MULTILINE)
```
rosiereflo, line 11: I would like to keep the contingency_event_time as it helps understanding  the regex, and also be resilient to the precision if not equals to 1e-3

rosiereflo (approved):

PR #589, opened 2023-02-20
#476 Tentative to disable the warning which requieres to use one argu…

barondim:
…ment with DYN::Error macro

### DIFF
--- a/sources/Inputs/src/Configuration.cpp
+++ b/sources/Inputs/src/Configuration.cpp
@@ -24,6 +24,8 @@
 #include <boost/property_tree/ptree.hpp>
 #include <limits>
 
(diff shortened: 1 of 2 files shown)


barondim:
I'm not sure it will work since I can't reproduce the compilation error locally.

sources/Inputs/src/Configuration.cpp
```diff
@@ -24,6 +24,8 @@
 #include <boost/property_tree/ptree.hpp>
 #include <limits>
 
+#pragma GCC diagnostic warning "-Wvarargs"
```
rosiereflo, line 27: Please remove the warning from the CMake

rosiereflo (approved):

rosiereflo (approved):

Issue #586, opened 2023-02-16
Properly initialize frozen property of VRemote and SVC


rosiereflo:
Follows https://github.com/dynawo/dynawo/issues/2791

cossartque:
@rosiereflo can we close this issue ?

rosiereflo:
No, I still need to port this development on v1.4 branch. I'll do it ASAP

PR #594, opened 2023-03-06
#586 properly initialize frozen property of VRRemote and SVCs + renaming

rosiereflo:
closes #568

### DIFF
--- a/sources/Algo/include/GeneratorDefinitionAlgorithm.h
+++ b/sources/Algo/include/GeneratorDefinitionAlgorithm.h
@@ -149,12 +149,13 @@ class GeneratorDefinition {
    * @param qmax maximum reactive power for the generator
    * @param pmin minimum active power for the generator
    * @param pmax maximum active power for the generator
+   * @param q reactive power of the generator
    * @param targetP target active power of the generator
    * @param regulatedBusId the Bus Id this generator is regulating
    * @param isNuclear true if the energy source of this generator is nuclear
    */
   GeneratorDefinition(const inputs::Generator::GeneratorId& genId, ModelType type, const inputs::Node::NodeId& nodeId,
-                      const std::vector<ReactiveCurvePoint>& curvePoints, double qmin, double qmax, double pmin, double pmax, double targetP,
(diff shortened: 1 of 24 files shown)


sources/Outputs/src/ParDynModel.cpp
```diff
@@ -108,13 +109,19 @@ ParDynModel::writeSVCParameterSet(const inputs::SettingDataBase::Set& set, const
       if (genDefinition.hasTransformer())
         new_set->addParameter(helper::buildParameter("secondaryVoltageControl_XTfoPu_" + std::to_string(idx) + "_",
                                                      (genDefinition.isNetwork()) ? constants::generatorNucXPuValue : constants::generatorXPuValue));
+
+      if (genDefinition.q < genDefinition.qmax && genDefinition.q > genDefinition.qmin) {
```
marcochiaramello, line 113: I'd align the codition as in the modelica model (check on volatge values also and <= >=)

sources/Algo/include/GeneratorDefinitionAlgorithm.h
```diff
@@ -196,13 +199,13 @@ class GeneratorDefinitionAlgorithm {
    * @brief Constructor
    *
    * @param gens generators list to update
-   * @param busesWithDynamicModel map of bus ids to a generator that regulates them
+   * @param busesRegulatedBySeveralGenerators map of bus ids to a generator that regulates them
```
marcochiaramello, line 202: thanks for renaming this!

marcochiaramello (changes requested):

sources/Outputs/src/ParDynModel.cpp
rosiereflo: Regarding the voltages, it is not possible as initially URef=URegulated=targetV of the node

marcochiaramello: I see, ok for me then! thx! we can still have <= and >=, I think

marcochiaramello (changes requested):

sources/Outputs/src/ParDynModel.cpp
marcochiaramello: never mind, it is correct as it is

marcochiaramello (approved):

PR #603, opened 2023-03-14
#516 Throw an error if there is a SVC in flat starting point mode

barondim:

### DIFF
--- a/etc/Dictionaries/DFLError_en_GB.dic
+++ b/etc/Dictionaries/DFLError_en_GB.dic
@@ -29,6 +29,7 @@ UnknownProperty               =     couldn't find the property %1% in assembling
 
 //------------------ Outputs -----------------------
 MissingGeneratorParameterInSettings = parameter %1% not found in settings file for generator %2%
+NoSVCInFlatStartingPointMode =        simulation of secondary voltage controls is not possible with 'FLAT' starting point mode
 //------------------ Main ---------------------------
 EnvVariableMissing           =     cannot find environnement variable %1% please check runtime environment
 NetworkFileNotFound          =     network file %1% does not exist
--- a/sources/Context.h
+++ b/sources/Context.h
@@ -119,6 +119,14 @@ class Context {
    */
   void execute();
 
+  /**
+   * @brief returns if the assembling data base contains one or more SVCs
+   * @returns true if the assembling data base contains one or more SVCs, false otherwise
+   */
+  bool dynamicDataBaseAssemblingContainsSVC() const {
+    return dynamicDataBaseManager_.assembling().containsSVC();
+  }
+
  private:
   /// @brief Slack node origin
   enum class SlackNodeOrigin {
--- a/sources/Inputs/include/AssemblingDataBase.h
+++ b/sources/Inputs/include/AssemblingDataBase.h
@@ -399,6 +399,14 @@ class AssemblingDataBase {
   }
 
   /**
+   * @brief determines if the assembling data base contains one or more SVCs
+   * @returns true if the assembling data base contains one or more SVCs, false otherwise
+   */
+  bool containsSVC() const {
+    return containsSVC_;
+  }
+
+  /**
    * @brief Retrieve a property with its id
    * @param id property id
    * @returns property with the given id, empty property if not found
@@ -418,7 +426,8 @@ class AssemblingDataBase {
   std::unordered_map<std::string, MultipleAssociation> multipleAssociations_;       ///< list of multiple associations
   std::unordered_map<std::string, std::string> generatorIdToSingleAssociationsId_;  ///< generator id associated to the single associations that contains it
   std::map<std::string, DynamicAutomaton> dynamicAutomatons_;                       ///< list of dynamic automatons
-  std::unordered_map<std::string, Property> properties_;                            ///< list of properties
+  bool containsSVC_;                                      ///< true if the assembling data base contains one or more SVC, false otherwise
+  std::unordered_map<std::string, Property> properties_;  ///< list of properties
 };
 
 }  // namespace inputs
--- a/sources/Inputs/src/AssemblingDataBase.cpp
+++ b/sources/Inputs/src/AssemblingDataBase.cpp
@@ -15,6 +15,7 @@
 
 #include "AssemblingDataBase.h"
 
+#include "Constants.h"
 #include "Log.h"
 
 #include <xml/sax/parser/ParserFactory.h>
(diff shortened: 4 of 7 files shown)


barondim:
- [ ] TU are still to be added

tests/algo/TestAlgoDynModel.cpp
```diff
@@ -38,7 +38,7 @@ TEST(TestAlgoDynModel, base) {
   using dfl::algo::DynamicModelDefinitions;
   using dfl::inputs::DynamicDataBaseManager;
 
-  DynamicDataBaseManager manager("res/setting.xml", "res/assembling.xml");
+  DynamicDataBaseManager manager("res/setting.xml", "res/assembling.xml", dfl::inputs::Configuration::StartingPointMode::WARM);
```
barondim, line 41: Do we have to input warm or flat by default ?

sources/Inputs/src/AssemblingDataBase.cpp
```diff
@@ -79,6 +79,15 @@ AssemblingDataBase::AssemblingDataBase(const boost::filesystem::path& assembling
   } catch (const xml::sax::parser::ParserException& e) {
     throw Error(DynModelFileReadError, assemblingFilePath.c_str(), e.what());
   }
+
+  if (startingPointMode == dfl::inputs::Configuration::StartingPointMode::FLAT) {
+    for (std::map<std::string, DynamicAutomaton>::const_iterator dynamicAutomatonsIt = dynamicAutomatons_.begin();
+         dynamicAutomatonsIt != dynamicAutomatons_.end(); ++dynamicAutomatonsIt) {
+      if (dynamicAutomatonsIt->second.lib == "SecondaryVoltageControlSimp") {
```
barondim, line 86: Is this the right way to detect the element is a SVC ?

rosiereflo: I would propose to add a method "SVCInAssembling" and then use it in Context or Configuration to make this test as it does not belong to the ddb

rosiereflo: > Is this the right way to detect the element is a SVC ?

There is a constant for the SVC library name

etc/Dictionaries/DFLError_en_GB.dic
```diff
@@ -29,6 +29,7 @@ UnknownProperty               =     couldn't find the property %1% in assembling
 
 //------------------ Outputs -----------------------
 MissingGeneratorParameterInSettings = parameter %1% not found in settings file for generator %2%
+NoSVCInFlatStartingPointMode =        cannot use a svc in flat starting point mode
```
rosiereflo, line 32: I would propose "simulation of secondary voltage controls is not possible with 'FLAT' starting point mode"

sources/Inputs/include/AssemblingDataBase.h
```diff
@@ -418,7 +426,8 @@ class AssemblingDataBase {
   std::unordered_map<std::string, MultipleAssociation> multipleAssociations_;       ///< list of multiple associations
   std::unordered_map<std::string, std::string> generatorIdToSingleAssociationsId_;  ///< generator id associated to the single associations that contains it
   std::map<std::string, DynamicAutomaton> dynamicAutomatons_;                       ///< list of dynamic automatons
-  std::unordered_map<std::string, Property> properties_;                            ///< list of properties
+  bool containsSVC_;                                      ///< true if the assembling data base contains one or more SVC, false otherwise
```
rosiereflo, line 429: comment alignment

sources/Inputs/src/AssemblingDataBase.cpp
```diff
@@ -79,6 +80,14 @@ AssemblingDataBase::AssemblingDataBase(const boost::filesystem::path& assembling
   } catch (const xml::sax::parser::ParserException& e) {
     throw Error(DynModelFileReadError, assemblingFilePath.c_str(), e.what());
   }
+
+  std::map<std::string, inputs::AssemblingDataBase::DynamicAutomaton>::const_iterator dynamicAutomatonsIt;
```
rosiereflo, line 84: Iterators should always be declared within the for loop

rosiereflo: Do we really need to loop afterward? maybe we could get this information on the fly inside the parser?

rosiereflo (changes requested):

sources/Inputs/include/AssemblingDataBase.h
barondim: The comments were aligned by clang-format

sources/Inputs/src/AssemblingDataBase.cpp
barondim: I'm not sure it's possible because xml::sax::parser::Parser is in libxml, a dynawo 3rdParty.

rosiereflo: I meant in AssemblingXmlDocument

marcochiaramello (changes requested):

sources/Context.cpp
```diff
@@ -60,6 +60,11 @@ Context::Context(const ContextDef& def, const inputs::Configuration& config) :
     algoResults_(new algo::AlgorithmsResults()),
     jobEntry_{},
     jobsEvents_{} {
+  if (config_.getStartingPointMode() == inputs::Configuration::StartingPointMode::FLAT &&
```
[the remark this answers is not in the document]
rosiereflo: The other location where we could do it is in main:!buildContext as we need both the assembling and configuration to be available. What do you think?

sources/Context.h
```diff
@@ -119,6 +119,14 @@ class Context {
    */
   void execute();
 
+  /**
+   * @brief returns if the assembling data base contains one or more SVCs
+   * @returns the attribute value
```
rosiereflo, line 124: wrong comment

sources/Inputs/include/AssemblingDataBase.h
```diff
@@ -398,6 +398,14 @@ class AssemblingDataBase {
     return dynamicAutomatons_;
   }
 
+  /**
+   * @brief determines if the assembling data base contains one or more SVCs
+   * @returns the attribute value
```
rosiereflo, line 403: wrong comment

rosiereflo (changes requested):

rosiereflo (approved):

PR #617, opened 2023-03-30
#27 windows port

### LINKED ISSUE #27: Windows compilation

valleefred:
Close #27 


valleefred:
This pull request depends on https://github.com/dynawo/dynawo-algorithms/pull/317
The checks cannot succeed until the pull request for the windows port of dynawo-algorithms has been merged into master!

rosiereflo (approved):

sources/Algo/include/DynModelDefinitionAlgorithm.h
```diff
@@ -134,7 +134,7 @@ struct DynamicModelDefinition {
  * @brief Definition of models
  */
 struct DynamicModelDefinitions {
-  std::unordered_map<DynamicModelDefinition::DynModelId, DynamicModelDefinition> models;      ///< models by dynamic model id
+  std::map<DynamicModelDefinition::DynModelId, DynamicModelDefinition> models;                ///< models by dynamic model id
```
gautierbureau, line 137: this change is due to the uuid instead of hash ?

valleefred: No, this change was made to be deterministic in unit tests when models are iterated because hash functions used in unordered_map are different from Linux to Windows.

The same non deterministic hash function was used to compute model id, which caused issues in unit tests, hence the use of a deterministic uuid function instead.

gautierbureau: ok what are the operation used on this container ? is there any lost of performance with this change ?

gautierbureau (approved):
If the change map to unordered_map is performance wise ok we can go on

sources/Algo/include/DynModelDefinitionAlgorithm.h
rosiereflo: I don't think so as it contains the number of automaton from the assembling file. Another way to do this would be to sort the container right before the dump

PR #645, opened 2023-04-26
#27 Fix gtest target for recent cmake.

gautierbureau:


gautierbureau:
This seems to work with 3 versions of cmake: 3.14, 3.16 and 3.19... I need @valleefred to confirm it works for Windows and then we are ok

rosiereflo (approved):

Issue #630, opened 2023-04-07
Two VRRemote with same name are added when a node is regulated by several generators and a HVDC with vscs


rosiereflo:
A single VRRemote connected to the generators and hvdc should be instantiated in this case

PR #667, opened 2023-06-09
#630 Two VRRemote with same name are added when a node is regulated by several generators and a HVDC with vscs

barondim:

### DIFF
--- a/sources/Outputs/include/Dyd.h
+++ b/sources/Outputs/include/Dyd.h
@@ -22,6 +22,7 @@
 #include "DydHvdc.h"
 #include "DydLoads.h"
 #include "DydSVarC.h"
+#include "DydVRRemote.h"
 #include "DynModelDefinitionAlgorithm.h"
 #include "DynamicDataBaseManager.h"
 #include "GeneratorDefinitionAlgorithm.h"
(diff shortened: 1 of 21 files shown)


sources/Outputs/src/DydVRRemote.cpp
```diff
@@ -0,0 +1,41 @@
...
+        regulatedBusIds.insert(busToGenerator.first);
+    for (const auto& busToVSC : vscBusVSCDefinitionsMap)
+        regulatedBusIds.insert(busToVSC.first);
+
+    for (const BusId& regulatedBusId : regulatedBusIds) {
+        std::string id = constants::modelSignalNQprefix_ + regulatedBusId;
+        boost::shared_ptr<dynamicdata::BlackBoxModel> blackBoxModelVRRemote =
+                                                        helper::buildBlackBox(id, "VRRemote", basename + ".par", id);
+        dynamicModelsToConnect->addModel(blackBoxModelVRRemote);
+        dynamicModelsToConnect->addConnect(id, "vrremote_URegulated", constants::networkModelName, regulatedBusId + "_U_value");
```
rosiereflo, line 36: the connections to the generators/hvdc are missing
image

rosiereflo (changes requested):

sources/Outputs/src/DydVRRemote.cpp
rosiereflo: Yes, this is the point of this development

sources/Context.h
```diff
@@ -146,7 +146,7 @@ class Context {
    * @param cbk the callback to register
    */
   void onNodeOnMainConnexComponent(ProcessNodeCallBackMainComponent &&cbk) {
-    callbacksMainConnexComponent_.push_back(std::forward<ProcessNodeCallBackMainComponent>(cbk));
```
rosiereflo, line 149: I'm quite sure we need to use forward here. Anyway this is not related so please revert this change.

sources/Outputs/include/Dyd.h
```diff
@@ -70,27 +71,39 @@ class Dyd {
...
+        dydSVarC_(new DydSVarC(svarcsDefs)),
+        dydHvdc_(new DydHvdc(hvdcDefinitions)),
+        dydGenerator_(new DydGenerator(gens)),
+        dydDynModel_(new DydDynModel(models, gens, loaddefs)),
+        dydVRRemote_(new DydVRRemote(gens, hvdcDefinitions)) {}
 
     std::string basename_;                                                                    ///< basename for file
     std::string filename_;                                                                    ///< filepath for file to write
     std::shared_ptr<inputs::Node> slackNode_;                                                 ///< slack node to use
-    const algo::HVDCLineDefinitions &hvdcDefinitions_;                                        ///< list of hvdc definitions
```
rosiereflo, line 86: wrong merge?

sources/Outputs/src/DydVRRemote.cpp
```diff
@@ -0,0 +1,92 @@
...
+      dynamicModelsToConnect->addMacroConnect(connection);
+    }
+  }
+}
+
+void
+DydVRRemote::writeConnect(boost::shared_ptr<dynamicdata::DynamicModelsCollection>& dynamicModelsToConnect) {
+  const std::string vrremoteNqValue("vrremote_NQ");
+  for (const auto& keyValue : hvdcDefinitions_.hvdcLines) {
+    algo::HVDCDefinition hvdcLine = keyValue.second;
```
rosiereflo, line 79: why 2 methods to create connections?

sources/Outputs/src/ParVRRemote.cpp
```diff
@@ -0,0 +1,66 @@
...
+      busesRegulatedBySeveralGeneratorsToFrozen[generator.regulatedBusId] = false;
+    }
+  }
+
+  // delete BusId already in busesRegulatedBySeveralGenerators_ to not generate VRRemotes twice
+  algo::HVDCLineDefinitions::BusVSCMap vscBusVSCDefinitionsMapWithoutBusDuplicates(vscBusVSCDefinitionsMap_);
+  for (const auto& keyValue : busesRegulatedBySeveralGenerators_) {
+    if (vscBusVSCDefinitionsMapWithoutBusDuplicates.find(keyValue.first) != vscBusVSCDefinitionsMapWithoutBusDuplicates.end()) {
+      algo::HVDCDefinition::BusId busId = static_cast<algo::HVDCDefinition::BusId>(keyValue.first);
+      vscBusVSCDefinitionsMapWithoutBusDuplicates.erase(busId);
```
rosiereflo, line 41: Why do we need an intermediate container? We could do the test on the fly below?

rosiereflo (changes requested):

sources/Context.h
barondim: ProcessNodeCallBackMainComponent&& isn't a universal reference. It's a rvalue reference so I thought std::move was more appropriate.

rosiereflo: yes it is a universal reference to a rvalue within a template so forward is required.

barondim: std::function<void(const std::shared_ptr<inputs::Node> &, std::shared_ptr<dfl::algo::AlgorithmsResults> &)> isn't a template, right ?

rosiereflo: we can discuss this later on case by case. Anyway this is not related to this development so please revert this in the context of this PR.

sources/Outputs/include/Dyd.h
barondim: This attribute is just never used

sources/Outputs/src/DydVRRemote.cpp
barondim: Currently, the connections creation is splited in 3 methods : 

writeMacroConnector(dynamicModelsToConnect);
writeMacroConnect(dynamicModelsToConnect);
writeConnect(dynamicModelsToConnect);

but I can merge them into one big method : writeAllConnections

rosiereflo: writeMacroConnector can be keep as it creates the macros, however the two others can be merged IMHO

sources/Outputs/src/ParVRRemote.cpp
barondim: We can but I thought it could be risky to alter directly vscBusVSCDefinitionsMap_ ^^'

rosiereflo: This is not my proposal. I am proposing to test if the bus was already implemented directly in the hvdc loop

barondim: Yes, you are right. I could be optimized.

sources/Outputs/include/ParVRRemote.h
```diff
@@ -0,0 +1,61 @@
...
+    generatorDefinitions_(generatorDefinitions),
+    busesRegulatedBySeveralGenerators_(busesRegulatedBySeveralGenerators),
+    vscBusVSCDefinitionsMap_(vscBusVSCDefinitionsMap) {}
+
+  /**
+   * @brief enrich the parameter set collection for VRRemote
+   *
+   * @param paramSetCollection parameter set collection to enrich
+   */
+  void writeVRRemotes(boost::shared_ptr<parameters::ParametersSetCollection>& paramSetCollection);
```
rosiereflo, line 52: Should be private IMHO

rosiereflo (changes requested):

sources/Outputs/include/ParVRRemote.h
barondim: If this method is private, we can't call it in Par::write()

tests/outputs/TestDyd.cpp
```diff
@@ -316,3 +316,56 @@ TEST(Dyd, writeLoad) {
...
+                                                                outputPath.generic_string(),
+                                                                {},
+                                                                {},
+                                                                node,
+                                                                hvdcDefs,
+                                                                busesRegulatedBySeveralGenerators,
+                                                                manager,
+                                                                noModels,
+                                                                {}));
+
```
rosiereflo, line 363: I don't see any unit test where there are both hvdc and generators regulated by a VRRemote?

barondim: In this unit test :

dydWriter.def_.dydVRRemote_->hvdcDefinitions_.vscBusVSCDefinitionsMap contains BUS_1 and BUS_3

dydWriter.def_.dydVRRemote_->busesRegulatedBySeveralGenerators_ contains BUS_1 and BUS_2

The unit test checks that only one Model_Signal_NQ is created for BUS_1 even if it's both regulated by a HVDC and a generator

rosiereflo (approved):

Issue #673, opened 2023-06-21
Dynawo output files are not generated when simulation fails


rosiereflo:
outputs files such as timeline, finalState, etc are not generated when the simulation fails whereas they are generated when running the equivalent simulation with dynawo core

PR #684, opened 2023-07-07
#673 Dynawo output files are not generated when simulation fails

barondim:


sources/Context.cpp
```diff
@@ -329,7 +329,12 @@ void Context::execute() {
     // For a power flow calculation it is ok to directly run here a single simulation
     auto simu = boost::make_shared<DYN::Simulation>(jobEntry_, simu_context, networkManager_.dataInterface());
     simu->init();
-    simu->simulate();
+    try {
+      simu->simulate();
```
rosiereflo, line 333: Some error type are missing. It should be similar to the equivalent section in DYNSImulationLauncher.cpp

rosiereflo (approved):

Issue #680, opened 2023-06-26
[SVC] when the U measurement node is not available, the SVC should be ignored 


rosiereflo:
This means that the generators models should not contain a Rpcl2

PR #689, opened 2023-07-19
#680 Remove RPCL from generators regulated by partially connected SVCs

barondim:

### DIFF
--- a/sources/Algo/CMakeLists.txt
+++ b/sources/Algo/CMakeLists.txt
@@ -10,6 +10,7 @@
 set(SOURCES
 src/ContingencyValidationAlgorithm.cpp
 src/DynModelDefinitionAlgorithm.cpp
+src/DynModelFilterAlgorithm.cpp
 src/GeneratorDefinitionAlgorithm.cpp
 src/HVDCDefinitionAlgorithm.cpp
 src/LineDefinitionAlgorithm.cpp
--- a/sources/Algo/include/DynModelFilterAlgorithm.h
+++ b/sources/Algo/include/DynModelFilterAlgorithm.h
@@ -1,0 +1,80 @@
+//
+// Copyright (c) 2023, RTE (http://www.rte-france.com)
+// See AUTHORS.txt
+// All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at http://mozilla.org/MPL/2.0/.
+// SPDX-License-Identifier: MPL-2.0
+//
+
+/**
+ * @file  DynModelFilterAlgorithm.h
+ *
+ * @brief Dynaflow launcher algorithms to filter model definitions with partial connectivity : header file
+ *
+ */
+
+#pragma once
+
+#include "AssemblingDataBase.h"
+#include "DynModelDefinitionAlgorithm.h"
+#include "GeneratorDefinitionAlgorithm.h"
+
+namespace dfl {
+namespace algo {
+
+/**
+ * @brief DynModelFilterAlgorithm definition to remove model definitions with partial connectivity for algorithm
+ */
+class DynModelFilterAlgorithm {
+ public:
+  /**
+   * @brief Constructor
+   *
+   * @param assembling assembling database
+   * @param generators the generators list to update
+   * @param dynamicModelsToFilter dynamic models by model id
+   */
+  DynModelFilterAlgorithm(const inputs::AssemblingDataBase& assembling,
+                          GeneratorDefinitionAlgorithm::Generators& generators,
+                          std::map<DynamicModelDefinition::DynModelId, DynamicModelDefinition>& dynamicModelsToFilter) :
+    assembling_(assembling),
+    generators_(generators),
+    dynamicModelsToFilter_(dynamicModelsToFilter) {}
+
+  /**
+   * @brief Remove partially connected SVC, RPCL in generators not connected to SVC and filter remaining partially
+   *        connected dynamic models
+   */
+  void filter();
+
+ private:
+  /**
+   * @brief remove RPCL in generators if SVC is not connected to a bus and then remove the related SVC
+   */
+  void removeRpclInGeneratorsAndSvcIfMissingConnexionToSvc();
+
+  /**
+   * @brief Check if a SVC is connected to a bus
+   *
+   * @param svcModel SVC model to check
+   *
+   * @returns true if the SVC is connected to a bus, false otherwise
+   */
+  bool checkIfSVCconnectedToUMeasurement(const algo::DynamicModelDefinition& svcModel) const;
+
+  /**
+   * @brief Filter partially connected dynamic models
+   *
+   * Remove model definitions with partial connectivity
+   */
+  void filterPartiallyConnectedDynamicModels();
+
+  const inputs::AssemblingDataBase& assembling_;                                         ///< assembling database
+  GeneratorDefinitionAlgorithm::Generators& generators_;                                 ///< the generators list to update
+  std::map<DynamicModelDefinition::DynModelId, DynamicModelDefinition>& dynamicModelsToFilter_;  ///< models by dynamic model id
+};
+
+}  // namespace algo
+}  // namespace dfl
--- a/sources/Algo/include/GeneratorDefinitionAlgorithm.h
+++ b/sources/Algo/include/GeneratorDefinitionAlgorithm.h
@@ -139,6 +139,11 @@ class GeneratorDefinition {
   }
 
   /**
+   * @brief change the model of the generator to remove RPCL
+   */
+  void removeRpclFromModel();
+
+  /**
    * @brief Constructor
    *
    * @param genId generator id
--- a/sources/Algo/src/DynModelFilterAlgorithm.cpp
+++ b/sources/Algo/src/DynModelFilterAlgorithm.cpp
@@ -1,0 +1,126 @@
+//
+// Copyright (c) 2023, RTE (http://www.rte-france.com)
+// See AUTHORS.txt
+// All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at http://mozilla.org/MPL/2.0/.
+// SPDX-License-Identifier: MPL-2.0
+//
+
+#include "DynModelFilterAlgorithm.h"
+#include "Constants.h"
+#include "Log.h"
+
+namespace dfl {
+namespace algo {
+
+void
+DynModelFilterAlgorithm::filter() {
+  removeRpclInGeneratorsAndSvcIfMissingConnexionToSvc();
+  filterPartiallyConnectedDynamicModels();
+}
+
+void
+DynModelFilterAlgorithm::removeRpclInGeneratorsAndSvcIfMissingConnexionToSvc() {
+  std::vector<DynamicModelDefinition::DynModelId> svcToRemove;
+  for (const auto& dynamicModel : dynamicModelsToFilter_) {
+    if (dynamicModel.second.lib == dfl::common::constants::svcModelName) {
+      bool isSVCconnectedToUMeasurement = checkIfSVCconnectedToUMeasurement(dynamicModel.second);
+      if (!isSVCconnectedToUMeasurement) {
+        svcToRemove.push_back(dynamicModel.second.id);
+        for (const DynamicModelDefinition::MacroConnection& generatorConnectedToSVC : dynamicModel.second.nodeConnections) {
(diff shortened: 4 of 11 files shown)


barondim:
Is it necessary to add a NRT which tests the added RPCL removal ?

sources/Context.cpp
```diff
@@ -167,6 +167,7 @@ void Context::filterPartiallyConnectedDynamicModels() {
       if (found == modelDef.nodeConnections.end()) {
         LOG(debug, ModelPartiallyConnected, automaton.second.id);
         dynamicModels_.models.erase(automaton.second.id);
+        removeRpclFromGenerators(automaton.second.macroConnects);
```
rosiereflo, line 170: This should be done only if the model is a SVC with a missing connection to a node

sources/Context.cpp
```diff
@@ -199,6 +200,29 @@ void Context::filterPartiallyConnectedDynamicModels() {
   }
 }
 
+void Context::removeRpclFromGenerators(const std::vector<inputs::AssemblingDataBase::MacroConnect>& automatonGeneratorMacroConnects) {
+  for (const auto &automatonGeneratorMacroConnect : automatonGeneratorMacroConnects) {
```
rosiereflo, line 204: I wonder if it wouldn't be better to do this in the Dynamic models/Generator algorithms

rosiereflo (changes requested):

sources/Context.cpp
barondim: Before we call Context::walkNodesMain(), dynamicModels_.models and generators_ are empty

That's why I thought it was better to call removeRpclFromGenerators in filterPartiallyConnectedDynamicModels()

rosiereflo: yes but it is not the role of context to define models, we need to think on how to do this properly

sources/Context.cpp
rosiereflo: still not solved

sources/Algo/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ src/HVDCDefinitionAlgorithm.cpp
 src/LineDefinitionAlgorithm.cpp
 src/LoadDefinitionAlgorithm.cpp
 src/MainConnexComponentAlgorithm.cpp
+src/RPCLDefinitionAlgorithm.cpp
```
rosiereflo, line 18: I would prefer to put all the filtering process in this model and to call this DynModelFilterALgorithm

sources/Algo/include/GeneratorDefinitionAlgorithm.h
```diff
@@ -138,6 +138,11 @@ class GeneratorDefinition {
     return model == ModelType::REMOTE_DIAGRAM_PQ_SIGNALN || model == ModelType::REMOTE_SIGNALN_INFINITE || model == ModelType::REMOTE_SIGNALN_RECTANGULAR;
   }
 
+  /**
+   * @brief change the model of the generator to remove RPCL
+   */
+  void removeRpcl();
```
rosiereflo, line 144: I would propose removeRpclFromModel

sources/Algo/include/RPCLDefinitionAlgorithm.h
```diff
@@ -0,0 +1,58 @@
...
+   * @param assembling assembling database
+   * @param generators the generators list to update
+   * @param models models by dynamic model id
+   */
+  RPCLDefinitionAlgorithm(const inputs::AssemblingDataBase& assembling,
+                          GeneratorDefinitionAlgorithm::Generators& generators,
+                          const std::map<DynamicModelDefinition::DynModelId, DynamicModelDefinition>& models) :
+    assembling_(assembling),
+    generators_(generators),
+    models_(models) {}
```
rosiereflo, line 44: we should be more specific in the name. dynamicModels?

sources/Algo/src/RPCLDefinitionAlgorithm.cpp
```diff
@@ -0,0 +1,66 @@
...
+RPCLDefinitionAlgorithm::removeRpclInGeneratorsIfMissingConnexionToSVC() {
+    const std::map<std::string, inputs::AssemblingDataBase::DynamicAutomaton>& automatonsConfig = assembling_.dynamicAutomatons();
+    for (const auto &automaton : automatonsConfig) {
+        if (models_.count(automaton.second.id) == 0) {
+            continue;
+        }
+
+        const DynamicModelDefinition& modelDef = models_.at(automaton.second.id);
+        for (const auto &macroConnect : automaton.second.macroConnects) {
+            auto found = std::find_if(modelDef.nodeConnections.begin(), modelDef.nodeConnections.end(),
```
rosiereflo, line 27: I was expecting a test on the ElementType for example to know if there is a node in the macroConnections, this is only testing that all connections provided in the assembling were instantiated

sources/Algo/src/RPCLDefinitionAlgorithm.cpp
```diff
@@ -0,0 +1,66 @@
...
+            [&macroConnect](const algo::DynamicModelDefinition::MacroConnection &macroConnection) {
+                return macroConnection.id == macroConnect.macroConnection;
+            });
+            if (found == modelDef.nodeConnections.end()) {
+                if (automaton.second.lib == dfl::common::constants::svcModelName) {
```
rosiereflo, line 32: If this is only for svc then lets put this test outside of the loop

sources/Algo/src/RPCLDefinitionAlgorithm.cpp
```diff
@@ -0,0 +1,66 @@
...
+                    for (const auto &automatonGeneratorMacroConnect : automaton.second.macroConnects) {
+                        std::string genId = automatonGeneratorMacroConnect.id;
+                        if (assembling_.isSingleAssociation(genId)) {
+                        const auto &assoc = assembling_.getSingleAssociation(genId);
```
rosiereflo, line 36: as we have access to the dynamicMOdel.nodeConnections I don't think we need to use the assembling

rosiereflo (changes requested):

sources/Common/include/Constants.h
```diff
@@ -28,6 +28,7 @@ namespace constants {
 const std::string diagramDirectorySuffix{"_Diagram"};              ///< Suffix for the diagram directory
 const std::string svcModelName{"SecondaryVoltageControlSimp"};     ///< name of the Secondary Voltage Controller model
 const std::string rpcl2PropertyName{"ReactivePowerControlLoop2"};  ///< name of the property to annote RPCL2 generators in assembling
+const std::string svcToUMeasurement{"SVCToUMeasurement"};          ///< name of the UMeasurement
```
rosiereflo, line 31: This can be named as we want (it could be called TOTO) so we need another way to detect this is the node connection

sources/Context.cpp
```diff
@@ -131,8 +132,12 @@ bool Context::process() {
   }
   walkNodesMain();
 
+  algo::DynModelFilterAlgorithm dynModelFilterAlgorithm(dynamicDataBaseManager_.assembling(),
+                                                        generators_,
+                                                        dynamicModels_.models);
+  dynModelFilterAlgorithm.removeRpclInGeneratorsAndSvcIfMissingConnexionToSvc();
```
rosiereflo, line 138: those 2 calls should be inside the filter

rosiereflo (changes requested):
unit tests are missing

sources/Context.cpp
barondim: You mean I should merge removeRpclInGeneratorsAndSvcIfMissingConnexionToSvc() and filterPartiallyConnectedDynamicModels() methods in the same method ?

rosiereflo: No i mean that the details of what is done in the filtering should be called inside DynModelFilterAlgorithm class

barondim: The 2 calls are DynModelFilterAlgorithm class methods

sources/Algo/include/DynModelFilterAlgorithm.h
```diff
@@ -0,0 +1,67 @@
...
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at http://mozilla.org/MPL/2.0/.
+// SPDX-License-Identifier: MPL-2.0
+//
+
+/**
+ * @file  DynModelFilterAlgorithm.h
+ *
+ * @brief Dynaflow launcher algorithms for RPCL header file
```
rosiereflo, line 14: wrong comment

sources/Algo/src/DynModelFilterAlgorithm.cpp
```diff
@@ -0,0 +1,121 @@
...
+    const std::vector<inputs::AssemblingDataBase::MacroConnect>& macroConnects = assemblingAutomatonIt->second.macroConnects;
+    for (const DynamicModelDefinition::MacroConnection& nodeConnection : svcModel.nodeConnections) {
+      auto macroConnectIt = std::find_if(macroConnects.begin(), macroConnects.end(),
+                                        [&nodeConnection](const inputs::AssemblingDataBase::MacroConnect& macroconnect) {
+                                          return nodeConnection.id == macroconnect.macroConnection;
+                                        });
+      if (macroConnectIt != macroConnects.end()) {
+        const std::string singleAssociationId = macroConnectIt->id;
+        const inputs::AssemblingDataBase::SingleAssociation& singleAssociation = assembling_.getSingleAssociation(singleAssociationId);
+        if (singleAssociation.bus.is_initialized()) {
```
rosiereflo, line 59: wouldn't it be fore efficient to find the bus connection in the assembling and then look for it in the nodeConnections?

rosiereflo (changes requested):

sources/Algo/include/DynModelFilterAlgorithm.h
rosiereflo: wrong comment

sources/Algo/src/DynModelFilterAlgorithm.cpp
```diff
@@ -0,0 +1,126 @@
...
+  } else {
+    throw Error(UnknownAutomaton, svcModel.id);
+  }
+  return isSVCconnectedToUMeasurement;
+}
+
+void
+DynModelFilterAlgorithm::filterPartiallyConnectedDynamicModels() {
+  const std::map<std::string, inputs::AssemblingDataBase::DynamicAutomaton>& automatonsConfig = assembling_.dynamicAutomatons();
+  for (const auto& automaton : automatonsConfig) {
```
rosiereflo, line 79: This is not supposed to be applied on SVCs

sources/Algo/src/DynModelFilterAlgorithm.cpp
```diff
@@ -0,0 +1,126 @@
...
+                                              return nodeConnection.id == macroConnect.macroConnection;
+                                            });
+        if (macroConnectIt != svcModel.nodeConnections.end()) {
+          isSVCconnectedToUMeasurement = true;
+          break;
+        }
+      }
+    }
+  } else {
+    throw Error(UnknownAutomaton, svcModel.id);
```
rosiereflo, line 71: Should never happen

sources/Algo/src/DynModelFilterAlgorithm.cpp
```diff
@@ -0,0 +1,126 @@
...
+                                        });
+          if (foundGen != generators_.end()) {
+            foundGen->removeRpclFromModel();
+          }
+        }
+      }
+    }
+  }
+  for (const DynamicModelDefinition::DynModelId& svcId : svcToRemove) {
+    dynamicModelsToFilter_.erase(svcId);
```
rosiereflo, line 46: Please add a debug log in this case

rosiereflo (changes requested):

sources/Algo/src/DynModelFilterAlgorithm.cpp
barondim: It may be not safe to remove this exception

sources/Algo/src/DynModelFilterAlgorithm.cpp
barondim: Initially, on the master branch, the Context::filterPartiallyConnectedDynamicModels() method dealt with SVCs. Why should it be different now ?

sources/Algo/src/DynModelFilterAlgorithm.cpp
rosiereflo: an assert should be enough: if this happens it means that we have "invented" an automaton that was not in the assembling

rosiereflo:
Please solve the conflict and fix the doxygen before merging

rosiereflo (changes requested):

rosiereflo:
Could you please squash the commits?

PR #728, opened 2024-02-02
#719 Clean error if input config file contains a setting file without an assembling file or contains an assembling file without a settting file

barondim:
… an assembling file or contains an assembling file without a settting file


etc/Dictionaries/DFLError_en_GB.dic
```diff
@@ -10,6 +10,8 @@
 //------------------ Inputs -------------------------
 ErrorConfigFileRead           =     error while reading configuration file: %1%
 StartingPointModeDoesntExist  =     starting point mode %1% doesn't exist
+SettingFileWithoutAssemblingFile =  %1% contains a setting file but not an assembling file
+AssemblingFileWithoutSettingFile =  %1% contains a assembling file but not an setting file
```
rosiereflo, line 14: typo: an assembling file but not a setting file

rosiereflo (approved):

PR #771, opened 2024-07-26
#770 Update following 3259 dynawo.

gautierbureau:


gautierbureau:
Windows is not working because dynawo-algos distrib and nrt are not working for over a month following CTC commits so the investigation will take some time so we can merge this one

Issue #766, opened 2024-07-22
[DDB] create a mandatory field in macroConnect of the assembling 


rosiereflo:
it would look like this:
```
 <dynamicAutomaton id="MyAutomaton" lib="...">
    <macroConnect macroConnection="Macro1" id="..."/>
    <macroConnect macroConnection="Macro2" id="..." mandatory="true"/>
  </dynamicAutomaton>

```
When this flag is not there, the macroConnection is not considered during the check that automatically remove partially connected automaton.

This will be useful for automaton that monitor several values as if only 1 value connexion is incoherent, we want to keep the automaton so that it monitors the others values.

PR #774, opened 2024-07-30
#766 creation of mandatory attribute in macroConnect element

barondim:

### DIFF
--- a/etc/xsd/assembling.xsd
+++ b/etc/xsd/assembling.xsd
@@ -125,5 +125,6 @@ SPDX-License-Identifier: MPL-2.0 -->
   <xs:complexType name="MacroConnect">
     <xs:attribute name="macroConnection" use="required" type="xs:string"/>
     <xs:attribute name="id" use="required" type="xs:string"/>
+    <xs:attribute name="mandatory" type="xs:boolean"/>
   </xs:complexType>
 </xs:schema>
--- a/etc/xsd/assembling.xsd
+++ b/etc/xsd/assembling.xsd
@@ -125,6 +125,6 @@ SPDX-License-Identifier: MPL-2.0 -->
   <xs:complexType name="MacroConnect">
     <xs:attribute name="macroConnection" use="required" type="xs:string"/>
     <xs:attribute name="id" use="required" type="xs:string"/>
-    <xs:attribute name="mandatory" type="xs:boolean"/>
+    <xs:attribute name="mandatory" use="optional" type="xs:boolean"/>
   </xs:complexType>
 </xs:schema>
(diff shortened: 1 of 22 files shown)


etc/xsd/assembling.xsd
```diff
@@ -125,5 +125,6 @@ SPDX-License-Identifier: MPL-2.0 -->
   <xs:complexType name="MacroConnect">
     <xs:attribute name="macroConnection" use="required" type="xs:string"/>
     <xs:attribute name="id" use="required" type="xs:string"/>
+    <xs:attribute name="mandatory" type="xs:boolean"/>
```
gaouizer, line 128: I am wondering if the `use="optional"` attribute would be necessary in that case?

gaouizer (changes requested):
three general comments:

- wouldn't be necessary to edit `DynModelFilterAlgorithm::filterPartiallyConnectedDynamicModels` in order to prevent partially connected automaton deletion?
- adding tests is necessary to make sure that these pieces of code work as expected  
- Linux build failure in the CI seems to be related to doxygen, this needs to be investigated

barondim:
- [ ] I made a first draft for this task but we need to confirm if it was the work that was expected
- [ ] Addition of unit tests

etc/xsd/assembling.xsd
rosiereflo: Hi, yes, the default value would be "false"

sources/Algo/src/DynModelFilterAlgorithm.cpp
```diff
@@ -110,13 +110,17 @@ DynModelFilterAlgorithm::filterPartiallyConnectedDynamicModels() {
       continue;
 
     for (const inputs::AssemblingDataBase::MacroConnect& macroConnect : automaton.second.macroConnects) {
-      auto found = std::find_if(
+      if (!macroConnect.mandatory) {
```
rosiereflo, line 113: this is the opoosite: if the connection is mandatory, perform the check

rosiereflo:
looks ok now, only ut and nrt are missing

sources/Inputs/include/AssemblingDataBase.h
```diff
@@ -100,8 +100,14 @@ class AssemblingDataBase {
    * @brief Macro connect XML element defining a macro connection of a dynamic model
    */
   struct MacroConnect {
+    /**
+     * @brief Default constructor
+     */
+    MacroConnect() : mandatory(false) {}
```
rosiereflo, line 106: default is true

rosiereflo (changes requested):

rosiereflo (approved):

gaouizer (approved):

PR #785, opened 2024-09-09
#783 properly exit all processes when N simulation fails in NSA

### LINKED ISSUE #783: [NSA] in multi-processing, when the N fails only the root process fails and the others hang

rosiereflo:
closes #783

### DIFF
--- a/tests/main_n_sa/res/contingencies_badN.json
+++ b/tests/main_n_sa/res/contingencies_badN.json
@@ -1,0 +1,231 @@
+{
+  "version": "1.0",
+  "name": "list",
+  "contingencies": [
+    {
+      "id": "load_contingency",
+      "elements": [
+        {
+          "id": "_047abdd7-c766-11e1-8775-005056c00008",
+          "type": "LOAD"
+        }
+      ]
+    },
+    {
+      "id": "load_contingency_bad_id",
+      "elements": [
+        {
+          "id": "_XXX",
+          "type": "LOAD"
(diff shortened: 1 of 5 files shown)


tests/main_n_sa/res/contingencies_badN.json
```diff
@@ -0,0 +1,231 @@
...
+      "id": "load_contingency",
+      "elements": [
+        {
+          "id": "_047abdd7-c766-11e1-8775-005056c00008",
+          "type": "LOAD"
+        }
+      ]
+    },
+    {
+      "id": "load_contingency_bad_id",
```
gaouizer, line 15: So this will make the N simulation fail (gracefully thanks to your dev) or the SA's?

gaouizer (approved):

tests/main_n_sa/res/contingencies_badN.json
rosiereflo: in NSA mode + multi-process, only the root process runs the N simulation, and the others processes wait (before starting the scenarios in a second time when the N results are avaialble). 
This issue fixes the behavior when the N simulation fails for some reason: in this case, initially the root process was launching an exception and exited peacefully...while the others processes were waiting indefinitely as the sync method was never executed by the root process.

rosiereflo: thus the nrt added is a testcase for which the N fails

PR #797, opened 2024-10-16
#791 When a contingency set contains non-applicable elements, filter

### LINKED ISSUE #791: [AS/NSA] when a component is missing for a contingency with more than 1 elements, the whole contingency is filtered
We should keep the contingency, just remove the disconnected element from the list

rosiereflo:
them and still run a simulation for the remaining ones closes #791


tests/algo/TestContingencyAlgo.cpp
```diff
@@ -149,7 +149,8 @@ TEST(ContingencyValidation, base) {
   expectedValidContingencies.emplace("hvdcline");
 
   for (auto validContingency : validContingencies.get()) {
-    auto expected = expectedValidContingencies.find(validContingency.get().id);
+    std::cout << "BUBU?! " << validContingency.id << std::endl;
```
gaouizer, line 152: is this meant to be displayed?

gaouizer (changes requested):

tests/algo/TestContingencyAlgo.cpp
rosiereflo: haha no I totally forgot this one. Let me remove it

rosiereflo: solved

gaouizer (approved):

Issue #784, opened 2024-09-09
Nightly release and release without mpi


gautierbureau:
Follow https://github.com/dynawo/dynawo-algorithms/issues/399

PR #801, opened 2024-10-31
#784 Nightly release and release without mpi

barondim:

### DIFF
--- a/sources/main.cpp
+++ b/sources/main.cpp
@@ -310,9 +310,7 @@ int main(int argc, char *argv[]) {
 
     // NSA: wait so that the steady state computation is over for everyone and the results written by root process
     DYNAlgorithms::multiprocessing::Context::sync();
-#ifdef _MPI_
     mpiContext.broadcast(successN);
-#endif
     // NSA: Every process has to fail if the N ran by the root process failed
     if (!successN)
       return EXIT_FAILURE;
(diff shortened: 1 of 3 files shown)


gautierbureau:
we don't need to do it for Linux release

barondim:
> we don't need to do it for Linux release

On the dyn-algo task https://github.com/dynawo/dynawo-algorithms/pull/406 we build dyn-algo with and without MPI. Isn't it coherent to do the same with dfl ?

gautierbureau (approved):

sources/main.cpp
```diff
@@ -310,7 +310,9 @@ int main(int argc, char *argv[]) {
 
     // NSA: wait so that the steady state computation is over for everyone and the results written by root process
     DYNAlgorithms::multiprocessing::Context::sync();
+#ifdef _MPI_
```
gautierbureau, line 313: should be encapsulated in the context I think

gautierbureau (changes requested):

barondim:
- [ ] to be merged with https://github.com/dynawo/dynawo-algorithms/pull/441

barondim:
- [ ] to be merged with https://github.com/dynawo/dynawo-algorithms/pull/442

gautierbureau (approved):

PR #809, opened 2024-11-29
#808 Update following 3452 OS (deadband on U/Q tresholds)

### LINKED ISSUE #808: Update following 3452 OS (deadband on U/Q tresholds)

rosiereflo:
closes #808

### DIFF
--- a/sources/Outputs/src/ParGenerator.cpp
+++ b/sources/Outputs/src/ParGenerator.cpp
@@ -142,6 +142,10 @@ boost::shared_ptr<parameters::MacroParameterSet> ParGenerator::buildGeneratorMac
     }
   }
 
+  if (!def.hasRpcl() || def.isUsingDiagram()) {
+    macroParameterSet->addParameter(helper::buildParameter("generator_QDeadBand", 0.0001));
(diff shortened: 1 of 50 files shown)


sources/Outputs/src/ParGenerator.cpp
```diff
@@ -142,6 +142,10 @@ boost::shared_ptr<parameters::MacroParameterSet> ParGenerator::buildGeneratorMac
     }
   }
 
+  if (!def.hasRpcl() || def.isUsingDiagram()) {
+    macroParameterSet->addParameter(helper::buildParameter("generator_QDeadBand", 0.0001));
+    macroParameterSet->addParameter(helper::buildParameter("generator_UDeadBand", 0.0001));
```
gaouizer, line 147: Is that a good practice to hard code this values here or shall we add them in the constants ?

gaouizer (changes requested):

sources/Outputs/src/ParGenerator.cpp
rosiereflo: it is only used in this class so I didn't create a shared constant, but I can if you think it is important

gaouizer (approved):

PR #820, opened 2024-12-20
#810 Add the possibility to zip the inputs and outputs for N simulation

barondim:


scripts/envDFL.sh
```diff
@@ -494,120 +494,221 @@ build_tests_coverage() {
...
-    export_preload
-    if [ "${DYNAWO_USE_MPI}" == "YES" ]; then
+      fi
       "$MPIRUN_PATH" -np $NBPROCS $DYNAFLOW_LAUNCHER_INSTALL_DIR/bin/DynaFlowLauncher --log-level $DYNAFLOW_LAUNCHER_LOG_LEVEL \
                                                                                       --network $2 \
                                                                                       --config $3 \
                                                                                       --contingencies $4
+    fi
+  else
+    if [ -n "$ZIP_ARCHIVE" ]; then
```
rosiereflo, line 585: Maybe we should add a check that the zip archive exists

rosiereflo (changes requested):
It seems that the N output will not be zipped. Also, the documentation and tests needs to be updated.

scripts/envDFL.sh
barondim: The check is made in the C++ code

rosiereflo: yes, but it is also the case for network, config and contingencies files and yet they are tested in the other branch. This is to be homogeneous

sources/Common/src/ZipErrorMessage.cpp
```diff
@@ -0,0 +1,33 @@
...
+// SPDX-License-Identifier: MPL-2.0
+//
+
+#include "ZipErrorMessage.h"
+
+#include <sstream>
+
+
+std::string formatZipErrorMessage(const zip::ZipException& e) {
+  std::stringstream zipErrorStream;
```
rosiereflo, line 17: Why overriding the messages sent by libzip? The general exception catch should be enough ?

sources/Context.cpp
```diff
@@ -320,6 +328,31 @@ void Context::execute() {
       throw;
     }
     simu->terminate();
+
+    std::ostringstream timelineStream;
+    simu->printTimeline(timelineStream);
+    const boost::filesystem::path timelineFilePath = simu->getTimelineOutputFile();
```
rosiereflo, line 334: No need to add new API in the core as we have the jobs somewhere

barondim: I'm not sure it's possible to get the constraints, lost equipments and finalStatte output file names from the jobEntry_ attribute

barondim:
- [ ] to be merged with https://github.com/dynawo/dynawo/pull/3537

sources/Common/src/ZipErrorMessage.cpp
barondim: The error message given by libzip is not always clear. When the zip::Error::LIBARCHIVE_INTERNAL_ERROR error code is thrown, the message is clear : ''Unrecognized archive format". But when the zip::Error::FILE_NOT_FOUND error code is thrown, only the name of the missing file is given. So the error message is not clear in that case.

sources/main.cpp
```diff
@@ -132,6 +141,31 @@ static void execSimulation(boost::shared_ptr<dfl::Context> context, dfl::inputs:
...
+
+    if (outputIsZip) {
+      const std::string programLogFileRelativePath = runtimeConfig.programName + ".log";
+      const std::string programLogFileAbsolutePath = createAbsolutePath(programLogFileRelativePath, outputPath.generic_string());
+      dfl::common::Log::addLogFileContentInMapData(programLogFileRelativePath, programLogFileAbsolutePath, mapOutputFilesData);
+
+      boost::shared_ptr<zip::ZipFile> archive = zip::ZipFileFactory::newInstance();
+      const std::string archivePath = createAbsolutePath("outputs.zip", outputPath.generic_string());
+      if (!runtimeConfig.contingenciesFilePath.empty())
+        archive = zip::ZipInputStream::read(archivePath);
```
barondim, line 159: Why is it needed to read the archive ?

sources/Common/src/Options.cpp
```diff
@@ -154,37 +123,13 @@ Options::parse(int argc, char* argv[]) {
...
     std::cerr << e.what() << std::endl;
     return Request::ERROR;
-  } catch (const std::exception& e) {
-    std::cerr << e.what() << std::endl;
-    return Request::ERROR;
-  }
-}
-
-bool
-Options::areRequiredFilesMissing(const po::variables_map& vm, const boost::shared_ptr<zip::ZipFile>& archive) const {
```
barondim: So we totally abandon this check ?

sources/main.cpp
rosiereflo: it is used when it was already created by dynawo-algorithms and we complete it

sources/Common/src/Options.cpp
rosiereflo: it is done later in the code when trying to load the files

PR #830, opened 2025-01-22
#810 Follow-up : Add the possibility to zip the inputs and outputs for N simulation

barondim:


scripts/diff.py
```diff
@@ -78,6 +78,8 @@ def get_argparser():
 
         outputiidm_result_paths = list()
         outputiidm_result_paths.append(os.path.realpath(os.path.join(results_root, "outputs/finalState/outputIIDM.xml")))
+        if options.output_zip:
+            outputiidm_result_paths.append(os.path.realpath(os.path.join(results_root, "output/outputs/finalState/outputIIDM.xml")))
```
rosiereflo, line 82: It would be cleaner to update result_root instead of adding if for every type of files

sources/main.cpp
```diff
@@ -204,7 +205,24 @@ int main(int argc, char *argv[]) {
...
   if (!runtimeConfig.zipArchivePath.empty()) {
     if (mpiContext.isRootProc()) {
-      boost::shared_ptr<zip::ZipFile> archive = zip::ZipInputStream::read(runtimeConfig.zipArchivePath);
+      boost::shared_ptr<zip::ZipFile> archive;
+      try {
+        archive = zip::ZipInputStream::read(runtimeConfig.zipArchivePath);
+      } catch (const zip::ZipException& e) {
+        switch (e.getErrorCode()) {
+          case zip::Error::Code::LIBARCHIVE_INTERNAL_ERROR:
+            std::cerr << e.what() << " for the " << runtimeConfig.zipArchivePath << " archive" << std::endl;
```
rosiereflo, line 214: Can we add this in the dictionary?

barondim: So we totally override the libzip error in this case ? We get rid of e.what() in the code.

rosiereflo: You can add something like "Zip error for file ... : e.what()" for example

rosiereflo: yes, I'm just proposing to use the dictionnary with e.what() as parameter of the message

rosiereflo (approved):
needs an update in french dictionary, otherwise all good

PR #871, opened 2025-05-13
#870 add an option to define the minimal time step for the solver

### LINKED ISSUE #870: [Config] Add an option to define solver minimal time step

rosiereflo:
closes #870

### DIFF
--- a/sources/Inputs/include/Configuration.h
+++ b/sources/Inputs/include/Configuration.h
@@ -51,9 +51,9 @@ class Configuration {
 
   /**
    * @brief performs sanity checks on the configuration
-   * 
+   *
    * @attention This method should be called after the constructor to verify that the configuration is properly set up.
-   * 
+   *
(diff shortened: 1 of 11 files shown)


sources/Inputs/include/Configuration.h
```diff
@@ -303,6 +310,7 @@ class Configuration {
   double stopTime_ = 100.;                                                           ///< stop time for simulation
   boost::optional<double> precision_;                                                ///< Precision of the simulation
   double timeStep_ = 10.;                                                            ///< maximum value of the solver timestep
+  double minTimeStep_ = 1.;                                                          ///< minimum value of the solver timestep
```
gaouizer, line 313: I did not rry it myself but what happens if the user enters a `minTimeStep` > `timeStep` ?... it is possible that there is no issue but I does not seem to me that we do any check. Do you think it worth it?

gaouizer (approved):

Issue #855, opened 2025-03-14
[zip mode] add an option to select outputs name


rosiereflo:
today the name outputs.zip is hardcoded

PR #876, opened 2025-05-23
#855 Add optionnal outputZipName in configFile

edouardChevreau:


sources/Inputs/src/Configuration.cpp
```diff
@@ -162,6 +162,28 @@ static void updateActivePowerCompensationValue(Configuration::ActivePowerCompens
...
+ * @param saMode true if simulation is in SA, false otherwise
+ * @param parameterValueModified a parameter key is added in this if the value
+ * was redefined in the configuration file
+ */
+static void updateFileNameValue(std::string &fileName, std::string fileExtension, const boost::property_tree::ptree &tree, const std::string &key,
+                                const bool saMode, std::unordered_set<std::string> &parameterValueModified) {
+  helper::updateValue(fileName, tree, key, saMode, parameterValueModified);
+
+  std::size_t found = fileName.find(fileExtension);
+  if (found != fileName.length() - fileExtension.length()) {
```
rosiereflo, line 182: This is to allow the user to only specify the zip name and not its extension?

edouardChevreau: Yes, this way both formats are correct (with or without .zip extension)

rosiereflo (approved):

rosiereflo (approved):

PR #895, opened 2025-09-24
#894 Add an option to enable timetable so that platforms can follow

### LINKED ISSUE #894: [Config] Add an option to enable timetable

rosiereflo:
simulation' progress
closes #894


sources/Outputs/src/Job.cpp
```diff
@@ -128,6 +128,13 @@ std::unique_ptr<job::OutputsEntry> Job::writeOutputs() const {
     output->setOutputsDirectory("outputs");
   }
 
+  auto timeTableStep = def_.configuration.timeTableStep();
```
gaouizer, line 131: why not `unsigned int` ?

gaouizer (approved):

PR #935, opened 2026-03-30
#933 Only count shunt from the main connex component

rosiereflo:

### DIFF
--- a/sources/Algo/include/ShuntDefinitionAlgorithm.h
+++ b/sources/Algo/include/ShuntDefinitionAlgorithm.h
@@ -17,6 +17,7 @@
 
 #pragma once
 
+#include "AlgorithmsResults.h"
 #include "NetworkManager.h"
 #include "Node.h"
 
@@ -49,8 +50,9 @@ class ShuntCounterAlgorithm {
    * The algorithm counts the number of shunts by voltage level
    *
    * @param node the node to process
+   * @param algoRes pointer to algorithms results class
    */
-  void operator()(const NodePtr& node);
+  void operator()(const NodePtr& node, std::shared_ptr<AlgorithmsResults> &algoRes);
 
  private:
   ShuntCounterDefinitions& shuntCounterDefs_;  ///< the counter definitions to update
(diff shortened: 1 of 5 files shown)


sources/Algo/include/ShuntDefinitionAlgorithm.h
```diff
@@ -49,8 +50,9 @@ class ShuntCounterAlgorithm {
    * The algorithm counts the number of shunts by voltage level
    *
    * @param node the node to process
+   * @param algoRes pointer to algorithms results class
    */
-  void operator()(const NodePtr& node);
+  void operator()(const NodePtr& node, std::shared_ptr<AlgorithmsResults> &algoRes);
```
gaouizer: I would recommend consistency in the positonning of `&`

sources/Algo/src/ShuntDefinitionAlgorithm.cpp
```diff
@@ -18,7 +18,7 @@ namespace algo {
 ShuntCounterAlgorithm::ShuntCounterAlgorithm(ShuntCounterDefinitions& defs) : shuntCounterDefs_(defs) {}
 
 void
-ShuntCounterAlgorithm::operator()(const NodePtr& node) {
+ShuntCounterAlgorithm::operator()(const NodePtr& node, std::shared_ptr<AlgorithmsResults> &) {
```
gaouizer: why not using ` std::shared_ptr<AlgorithmsResults> & /* algoRes */` as in the rest of the project. Maybe more readable too.

sources/Inputs/src/NetworkManager.cpp
```diff
@@ -178,7 +178,7 @@ void NetworkManager::buildTree() {
         auto bus1 = sw->getBusInterface1();
         auto bus2 = sw->getBusInterface2();
         auto found1 = shuntsMap.find(bus1->getID());
-        auto found2 = shuntsMap.find(bus1->getID());
+        auto found2 = shuntsMap.find(bus2->getID());
```
gaouizer, line 181: good catch

gaouizer (changes requested):

sources/Algo/include/ShuntDefinitionAlgorithm.h
rosiereflo: solved

sources/Algo/src/ShuntDefinitionAlgorithm.cpp
rosiereflo: solved

gaouizer (approved):

PR #942, opened 2026-04-13
#940 Use Rpcl models when a generator with a remote voltage regulation is in a secondary voltage regulation area

### LINKED ISSUE #940: Better handle when remote voltage regulation and secondary voltage control are used together
The secondary voltage control should take over the remote voltage control

rosiereflo:

### DIFF
--- a/sources/Algo/include/GeneratorDefinitionAlgorithm.h
+++ b/sources/Algo/include/GeneratorDefinitionAlgorithm.h
@@ -249,6 +249,18 @@ class GeneratorDefinitionAlgorithm {
    */
   bool isDiagramRectangular(const inputs::Generator &generator) const;
 
+  /**
+   * @brief selection of the most common generator models
+   *
+   * @param generator reference to input generator
+   * @param withTfo whether the transformer should be modeled within the model used or not
+   * @param infiniteReactiveLimits whether the selected generator should model reactive limits
+   * @param isInSVC whether the target generator belongs to a secondary voltage control area
+   * @param isRPCL2 whether the target generator uses reactive power control loop 2
+   * @return the selected generator model
+   */
+  ModelType selectDiagramGenerators(const inputs::Generator &generator, bool withTfo, bool infiniteReactiveLimits, bool isInSVC, bool isRPCL2) const;
+
   Generators &generators_;                                                        ///< the generators list to update
   const inputs::NetworkManager::BusMapRegulating &busesToNumberOfRegulationMap_;  ///< mapping of busId and the number of generators that regulates them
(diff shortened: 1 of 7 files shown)


sources/Outputs/src/ParDynModel.cpp
```diff
@@ -127,7 +127,6 @@ std::shared_ptr<parameters::ParametersSet> ParDynModel::writeSVCParameterSet(con
         assert(genInitialParamToValues.find("Qr_" + std::to_string(it->second)) != genInitialParamToValues.end());
         new_set->addParameter(helper::buildParameter("Qr_" + std::to_string(idx), genInitialParamToValues["Qr_" + std::to_string(it->second)]));
       }
-      //      new_set->addParameter(helper::buildParameter("Participate0_" + std::to_string(idx), true));
```
gaouizer, line 130: Consider removing this comment ?

sources/Algo/include/GeneratorDefinitionAlgorithm.h
```diff
@@ -249,6 +249,18 @@ class GeneratorDefinitionAlgorithm {
...
+   * @brief selection of the most common generator models
+   *
+   * @param generator reference to input generator
+   * @param withTfo whether the transformer should be modeled within the model used or not
+   * @param infiniteReactiveLimits whether the selected generator should model reactive limits
+   * @param isInSVC whether the target generator belongs to a secondary voltage control area
+   * @param isRPCL2 whether the target generator uses reactive power control loop 2
+   * @return the selected generator model
+   */
+  ModelType selectDiagramGenerators(const inputs::Generator &generator, bool withTfo, bool infiniteReactiveLimits, bool isInSVC, bool isRPCL2) const;
```
gaouizer, line 262: It looks like part of the generator model's selection logic has been moved to this functions (obviously the most redundant one). But not all of the logic since the `operator()` some times returns the result of `selectDiagramGenerators` and some other times directly a `ModelType`. I assume you went as far as reasonably possible in to this logic mooving? (just a comment)

gaouizer (changes requested):
One parasit commented to be removed

sources/Outputs/src/ParDynModel.cpp
rosiereflo: this is exactly what is done :)

sources/Algo/include/GeneratorDefinitionAlgorithm.h
rosiereflo: yes, the branch which stayed outside are the one that are unique

gaouizer (approved):

PR #946, opened 2026-04-15
#943 Connections with IntPin updated after OS 3748

### LINKED ISSUE #943: Update connections with IntPin after OS #3748

ErwanGuichard49:
closes [#943](https://github.com/dynawo/dynaflow-launcher/issues/943)

### DIFF
--- a/tests/outputs/reference/TestDydEvent/TestDydEvent.dyd
+++ b/tests/outputs/reference/TestDydEvent/TestDydEvent.dyd
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <dyn:dynamicModelsArchitecture xmlns:dyn="http://www.rte-france.com/dynawo">
   <dyn:macroConnector id="MC_EventQuadripoleDisconnection">
-    <dyn:connect var1="event_state1_value" var2="@NAME@_state_value"/>
+    <dyn:connect var1="event_state1" var2="@NAME@_state_value"/>
   </dyn:macroConnector>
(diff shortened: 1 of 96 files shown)


tests/outputs/reference/TestDydEvent/TestDydEvent.dyd
joyelfeghali, line 1: I think you should also modify tests/algo/res/assembling.xml file : 
`<connection var1="currentLimitAutomaton_order" var2="@NAME@_state_value" required="true" internal="false"/>`

joyelfeghali (changes requested):

tests/outputs/reference/TestDydEvent/TestDydEvent.dyd
ErwanGuichard49: You are right. I wonder whether this file is part of the non-regression tests.

joyelfeghali: If it is like the one in dynawo-interface,  I guess not

ErwanGuichard49: According to @rosiereflo, the assembling.xml file is not tested for simulation but for C++ only. I can update this file but it will have no impact.

joyelfeghali: Okay I think it is better if it is updated

ErwanGuichard49: Done.

joyelfeghali (approved):

PR #961, opened 2026-06-22
#959 add criteria files

gaouizer:

- [x] main documentation was updated (update of input/output file)


etc/Dictionaries/DFLError_en_GB.dic
```diff
@@ -16,6 +16,7 @@ InvalidActivePowerCompensation =    parameter ActivePowerCompensation can't be s
 ChosenOutputDoesntExist       =     chosen output %1% doesn't exist
 NoFlatStartingPointModeInSA   =     flat mode cannot be used in security analysis
 StartingDumpFileNotFound      =     an starting dump file %1% was specified in the configuration file but it does not exist
+CriteriaFileNotFound          =     a criteria file %1% was specified in the configuration file but it does not exist
```
rosiereflo, line 19: The french dictionary on the private side will need an update as well

sources/Inputs/include/Configuration.h
```diff
@@ -341,6 +347,7 @@ class Configuration {
   // SA
   double timeOfEvent_ = 10.;                                ///< time for contingency simulation (security analysis only)
   boost::filesystem::path startingDumpFilePath_;            ///< starting dump file path
+  boost::filesystem::path criteriaFilePath_;                ///< criteria file path (optional)
                                                             ///< are already described in the static description
```
rosiereflo, line 350: I think this comment should be above the new attribute

rosiereflo (changes requested):

etc/Dictionaries/DFLError_en_GB.dic
gaouizer: indeed

rosiereflo (approved):

